Issue #8, opened 2023-08-29
Add pagination


dfarr:
APIs that require pagination
- SearchPromises
- ReadSubscriptions

dfarr:
# Stable pagination

If we plan to support stable pagination we can rule out offset based pagination right away. Cursor based pagination could be based on a unique column that is known to only increase (or decrease) over time. At first glance, one of our timestamp columns seem like a good candidate.

Promises have two timestamps:
- created_on
- completed_on

However, using these columns would prove problematic for two reasons:
1. Both values are set to the tick time and therefore often not unique (a batch of promises are created/completed on each tick).
2. The tick time is not guaranteed to always increase.

The second point is especially tricky. While it is true that the tick time is guaranteed to only increase within a single process, what happens if my application restarts? What happens if I scale up my application and one of my instances clock is behind the other?

Instead, we can opt to add an auto incrementing column to our table and use this in our query, this is sometimes called keyset based pagination.
```sql
SELECT
  *
FROM
  promises
WHERE
  auto_incrementing_id < $1
ORDER BY
  auto_incrementing_id
LIMIT
  $2
```

# Non stable pagination

Easier to implement, possibly more performant. Can our client tolerate duplicate results?

dtornow:
Requirements for pagination:

- a promise may be included twice
- a promise must *not* be omitted

dfarr:
Closed by #71 and #66

PR #66, opened 2023-09-22
Enable pagination for search api

dfarr:
Adds an auto incremented column named `sort_id` to our `promises` table. The search query sorts results by this column.

The SearchPromises coroutine resolves the request with a cursor containing the request to get the next page. All parameters are set to the same value as the initial request with the exception of `SortId` (initially `nil`) which is set to the last seen id.

```go
&types.Response{
	Kind: types.SearchPromises,
	SearchPromises: &types.SearchPromisesResponse{
		Status:   types.ResponseOK,
		Promises: result.promises,
		Cursor: &types.Cursor[types.SearchPromisesRequest]{
			Next: &types.SearchPromisesRequest{
				Q:      req.SearchPromises.Q,
				T:      req.SearchPromises.T,
				States: req.SearchPromises.States,
				Limit:  req.SearchPromises.Limit,
				SortId: &result.LastSortId,
			},
		},
	},
}
```

The `Cursor` struct implements custom json marshaling logic to serialize this structure as a JWT token, this way when we get the cursor back in subsequent requests we can verify the cursor has not been tampered with.

```go
type Cursor[T any] struct {
	Next *T
}

func (c *Cursor[T]) MarshalJSON() ([]byte, error) {
	// returns a serialized JWT token containing the next request T in the claims
}

func (c *Cursor[T]) UnmarshalJSON(data []byte) error {
	// deserializes and verifies a JWT token, sets the next request T to the value from the claims
}
```

Check out the implementation [here](https://github.com/resonatehq/resonate/pull/66/files#diff-3c2af47ef3959855a6e3b3b1d38c411821b5d40c4545959948088a7704c29271).


dfarr:
Closes #8

Issue #19, opened 2023-09-06
Create dsl for search API


dfarr:
Let's start with the most bare bones dsl, either:
- wildcard only with `*`
- prefix only search

dfarr:
Once this is in place, we can reinstate SearchPromises API in DST.

## DST

Deterministic simulation testing (DST) is fundamental to our development strategy at Resonate. Every four hours we run four iterations of DST given some seed value `s`.

1. A run with `seed=s` and `store=sqlite`, let's call this `s1`.
2. A run with `seed=s` and `store=sqlite`, let's call this `s2`.
3. A run with `seed=s` and `store=postgres`,  let's call this `p1`.
4. A run with `seed=s` and `store=postgres`, let's call this `p2`.

Because our system is deterministic, that is, two systems that receive the same requests in the same order are **guarenteed** to be in the same state, we can now perform the following three diffs:

1. diff of `s1` and `s2`
2. diff of `p1` and `p2`
3. diff of `s1` and `p1`

If any of the diffs fail, we know that we have broken our guarantee of determinism and have strong evidence that there is something more fundamentally incorrect that needs to be addressed in our codebase. See [#104](https://github.com/resonatehq/resonate/actions/runs/6225277140) for an example DST workflow run, download the logs and see for yourself if there is any difference!

Here is an example snippet from `s1` starting at line `15616`:
```
level=INFO msg=DST t=49 req="CreateSubscription(id=305, promiseId=379, url=https://resonatehq.io/408)" res="CreateSubscription(status=201, subscriptions=Subscription(id=305, promiseId=379, url=https://resonatehq.io/408, retryPolicy=RetryPolicy(delay=406, attempts=2)))" err=<nil>
level=INFO msg=DST t=49 req="ReadPromise(id=487)" res="ReadPromise(status=200, promise=Promise(id=487, state=PENDING, param=Value(ikey=<nil>, data=269), value=Value(ikey=<nil>, data=), timeout=386))" err=<nil>
level=INFO msg=DST t=49 req="RejectPromise(id=559, ikey=<nil>)" res="RejectPromise(status=404, promise=<nil>)" err=<nil>
level=INFO msg=DST t=49 req="DeleteSubscription(id=29, promiseId=315)" res="DeleteSubscription(status=404)" err=<nil>
[...]

PR #72, opened 2023-09-28
Add search dsl

dfarr:
Currently only simple wildcard queries with an asterisk `*` are supported

| query | sql | regex equivalent |
| --- | --- | --- |
| `*` | `WHERE id LIKE %` | `^.*$` |
| `foo.*` | `WHERE id LIKE foo.%` | `^foo\..*$` |
| `*.bar` | `WHERE id LIKE %.bar` | `^.*bar\.$` |


dfarr:
Closes #19

PR #82, opened 2023-10-10
refactor common code from http and grpc subsystem into a service

huesflash:
Pros:
- This will help keep common code in one place
- flexibility for wrapping service code with various frameworks (ex: gin, fiber) 
- existing api is maintained 

Cons:
 - some extra copies for translating protbufs into request arguments
 - naming isn't great

### DIFF
--- a/internal/app/subsystems/api/grpc/grpc.go
+++ b/internal/app/subsystems/api/grpc/grpc.go
@@ -25,7 +25,7 @@ type Grpc struct {
 }
 
 func New(api api.API, config *Config) api.Subsystem {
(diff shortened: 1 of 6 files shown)


dfarr:
🎉 🎉 Thank you so much @huesflash, your PR is the first open contribution to resonate!!!

I love the idea to factor out the common code into a delegated service. It also makes a lot of sense if (when) we decide to support additional protocols.

dfarr (approved):

huesflash:
Added the missing service test!
Should help with coverage as well

dfarr:
Thank you @huesflash !

PR #93, opened 2023-11-02
refactor(flags): keep flags consistent and clean with just ports

guergabo:
**Before**
Inconsistent input and log outputs and unnecessarily verbose. 
Screenshot 2023-11-02 at 10 22 11 AM

**After**
Consistent input and output for all 3 servers. Also consistent with db flags which take in ports. 
Screenshot 2023-11-02 at 10 25 30 AM


guergabo:
spoke to @dfarr. closing for now.

PR #107, opened 2023-11-09
fix(race): fixing race condition found with test harness

guergabo:
**Changes**
- Add a buffered channel of size 1 to get rid of race condition. 
- Update the callback to just write to the channel since a write to a closed channel already causes a panic. 

**Context**
I made the assumption the `sendOrPanic` only cares that the api goroutine didn't die and is available to complete the request. Since we defer the close() in the api goroutine, it means that if the kernel can write to the channel (without panicing), the api layer goroutine is available and has not exited or paniced.  so should be same behavior as before minus race condition.

### DIFF
--- a/internal/app/subsystems/api/service/service.go
+++ b/internal/app/subsystems/api/service/service.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/resonatehq/resonate/internal/api"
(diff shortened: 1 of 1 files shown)


internal/app/subsystems/api/service/service.go
```diff
@@ -268,10 +268,6 @@ func (s *Service) sendOrPanic(cq chan *bus.CQE[t_api.Request, t_api.Response]) f
 			Error:      err,
 		}
 
-		select {
-		case cq <- cqe:
-		default:
-			panic("response channel must not block")
```
dfarr, line 274: I would be inclined to leave the panic, for example, what happens if a send occurs again before the function completes?

dfarr (approved):

PR #110, opened 2023-11-13
feat(flags): expose subscriptions api behind feature flag for test harness

### LINKED ISSUE #109: add feature flagging
Our primary api, the promise api, should be the only one exposed at this time imo. All other apis such as subscriptions should be hidden until they are ready for public consumption. 

I propose that we hide the experimental features behind feature flags. Users can enable the experimental features if they want, but they explicitly know not to expect a mature feature.

Looked at: 
- https://prometheus.io/docs/prometheus/latest/feature_flags/
- https://gofeatureflag.org/docs 
- https://github.com/open-feature

## Alternatives you've considered
Another alternative would be to just document them as experimental/unfinished features. Although, we should prob do both. 

## Additional context
n/a

guergabo:
Fixes https://github.com/resonatehq/resonate/issues/109

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -6,6 +6,7 @@ import (
 	netHttp "net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
@@ -162,6 +163,7 @@ var serveCmd = &cobra.Command{
 
 func init() {
 	// api
+	serveCmd.Flags().Var(&ExperimentalFeatures{}, "api-enable-feature", "enable features that are disabled by default since they are considered experimental")
 	serveCmd.Flags().Int("api-size", 100, "size of the submission queue buffered channel")
 	serveCmd.Flags().String("api-http-addr", "0.0.0.0:8001", "http server address")
 	serveCmd.Flags().Duration("api-http-timeout", 10*time.Second, "http server graceful shutdown timeout")
@@ -170,6 +172,7 @@ func init() {
 	_ = viper.BindPFlag("api.size", serveCmd.Flags().Lookup("api-size"))
 	_ = viper.BindPFlag("api.subsystems.http.addr", serveCmd.Flags().Lookup("api-http-addr"))
 	_ = viper.BindPFlag("api.subsystems.http.timeout", serveCmd.Flags().Lookup("api-http-timeout"))
+	_ = viper.BindPFlag("api.subsystems.http.enable", serveCmd.Flags().Lookup("api-enable-feature"))
 	_ = viper.BindPFlag("api.subsystems.grpc.addr", serveCmd.Flags().Lookup("api-grpc-addr"))
 
 	// aio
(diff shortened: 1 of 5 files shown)


guergabo:
remove api

cmd/serve.go
```diff
@@ -162,6 +163,7 @@ var serveCmd = &cobra.Command{
 
 func init() {
 	// api
+	serveCmd.Flags().Var(&ExperimentalFeatures{}, "enable-feature", "enable features that are disabled by default since they are considered experimental")
```
dfarr, line 166:
```suggestion
	serveCmd.Flags().Var(&ExperimentalFeatures{}, "enable-feature", "enable experimental features that are disabled by default")
```

cmd/serve.go
```diff
@@ -226,3 +229,28 @@ func init() {
 	serveCmd.Flags().SortFlags = false
 	rootCmd.AddCommand(serveCmd)
 }
+
+type ExperimentalFeatures []string
```
dfarr, line 233: could this move to the `flags.go` file?

internal/app/subsystems/api/http/http.go
```diff
@@ -39,6 +42,13 @@ func New(api api.API, config *Config) api.Subsystem {
 	r.POST("/promises/:id/resolve", s.resolvePromise)
 	r.POST("/promises/:id/reject", s.rejectPromise)
 
+	// Subscription API (experimental)
+	if slices.Contains(config.Enable, "subscription") {
```
dfarr, line 46:
```suggestion
	if slices.Contains(config.Enable, "subscriptions") {
```

internal/app/subsystems/api/service/request.go
```diff
@@ -58,3 +63,34 @@ type RejectPromiseHeader struct {
 type RejectPromiseBody struct {
 	Value promise.Value `json:"value"`
 }
+
+// Subscription
+
+type SearchSubscriptionsParams struct {
```
dfarr, line 69: We probably won't have a search subscriptions API, just a list all subscriptions on a promise (which is also paginated),

internal/app/subsystems/api/service/request.go
```diff
@@ -58,3 +63,34 @@ type RejectPromiseHeader struct {
...
+	// Q      string `form:"q" json:"q"`
+	PromiseId string `json:"promiseId"`
+	Limit     int    `form:"limit" json:"limit"`
+	// Cursor    string `form:"cursor" json:"cursor"`
+}
+
+type CreateSubscriptionHeader struct {
+	RequestId      string                  `header:"request-id"` // qq: just for grpc?
```
dfarr, line 77: This is an optional parameter for tracking requests cross-systems. If provided this field is used in corresponding logs, if not provided I believe we generate a uuid for each request.

dfarr: you can set in both http/grpc

PR #111, opened 2023-11-14
feat(error): init error handling philosophy

### LINKED ISSUE #95: Standardize error messaging
This is an enhancement for both the implementation and the [Durable Promise API Spec](https://github.com/resonatehq/durable-promise). 

Currently, for 404 errors we are returning nil. While the status code indicates the type of failure we should come up with a error object standard to include clear and helpful messaging to consistently add to all the endpoints. 

Screenshot 2023-11-04 at 6 16 58 PM

### LINKED ISSUE #96: Do not return resource for anything other than 2xxs
The below image shows that if i DO NOT use an idempotency key and yet try to create the same object twice i get back the same promise. It looks idempotent on the CLI. The server does return different status codes for both calls 201 and then 403. But, this should be an error message. 

Also, I believe 409 makes more sense for this type of error since its more of a resource conflict than a forbidden request. 

screenshot_2023-11-03_at_1 41 11_pm

### LINKED ISSUE #98: Do not implicitly set timeout to 0
If the end user does not set the timeout, then there should just not be any timeout. Currently, if an end user does not set a timeout it implicitly sets it to 0 and the promise immediately goes into a `REJECTED_TIMEDOUT` state.  

Screenshot 2023-11-04 at 6 29 12 PM

### LINKED ISSUE #99: Consistent status codes with Idempotency keys
This is a bit of a suggestion, but also a question. Currently, if I create a promise with an `idempotency-key` I get a 201 status code for the first request. Any subsequent requests get a 200 status code. 

This is probably not technically wrong, but should the server not return the same 201 always? If not, we expect that the client-side handles both status codes correctly and kinda defeats the usability aspect of an `idempotency-key` imo.  

Any value from differentiating the initial call to the subsequent calls in the status code?

guergabo:
### Changes
Fixes https://github.com/resonatehq/resonate/issues/95
Fixes https://github.com/resonatehq/resonate/issues/96
Fixes https://github.com/resonatehq/resonate/issues/98
Fixes https://github.com/resonatehq/resonate/issues/99

### Context

**Error Duality**

Expected Errors 
1. `application level errors` (tolerable, expected, non-retryable) -> ResponseStatus && validation layer using json tags -- protobuf

2. `platform level errors` (tolerable, expected, retryable) -> errors.New() then convert to ResponseStatus at api layer

Unexpected Errors
1. `catastrophic` (intolerable, unexpected) -> panic()


internal/api/errors.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+	default:
+		panic(fmt.Sprintf("unknown status code %d", status))
+	}
+
+	// add url to error code page -- think about this
+	apiError.Details = append(apiError.Details, ErrorDetail{
+		Metadata: map[string]string{
+			"url": fmt.Sprintf("https://docs.resonatehq.io/reference/error-codes/%d", apiError.Code), // TODO: make this the internal error code
```
dfarr: This is amazing!!!

guergabo:
@vaibhawvipul let's talk tmrrw about this pr. this changes a bit of the expected api behavior. for example introducing 409 status code and returning error messages for unsuccessful requests instead of nil or promises.  the sdk will need to adjust a bit. additionally, the client deciding which errors are retryable becomes trivial once this is merged. 

For grpc, we return proper grpc error codes now and status in response is only to capture dedup logic (a.k.a, did this request change something in the database or not).

internal/kernel/t_api/status.go
```diff
@@ -0,0 +1,92 @@
...
+	ErrFailedToParseSubscriptionRecord
+	ErrFailedToDeleteSubscription
+)
+
+type ResonateError struct {
+	code     ResonateErrorCode
+	metadata string
+}
+
+func NewResonateError(code ResonateErrorCode, metadata string) *ResonateError {
```
dfarr: How would you feel about changing metadata to reason?

internal/app/coroutines/cancelPromise.go
```diff
@@ -181,3 +184,19 @@ func CancelPromise(metadata *metadata.Metadata, req *t_api.Request, res func(*t_
 		}
 	})
 }
+
+// move to utils
+func getForbiddenStatus(currentState promise.State) t_api.ResponseStatus {
```
dfarr: Let's move this to utils ;)

internal/api/errors.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// Domain is the domain of the error.
+	Domain string `json:"domain,omitempty"`
+
+	// Metadata is additional metadata about the error.
+	Metadata map[string]string `json:"metadata,omitempty"`
+}
+
+// HandlePlatformLevelError handles platform level errors and returns an APIErrorResponse.
+func HandlePlatformLevelError(err *t_api.ResonateError) *APIErrorResponse {
```
dfarr: Should we rename this HandleResonateError to match?

internal/api/errors.go
```diff
@@ -0,0 +1,299 @@
...
+			Message: "A promise with this identifier already exists",
+			Status:  t_api.StatusPromiseAlreadyExists.String(),
+		}
+
+	default:
+		panic(fmt.Sprintf("unknown status code %d", status))
+	}
+
+	apiError.Details = append(apiError.Details, ErrorDetail{
+		Type:    "ApplicationLevelError",
```
dfarr: I'm not sure "application level" is the best way to communicate with the client, any other ideas for what we could call this class of errors?

internal/app/coroutines/notifiySubscriptions.go
```diff
@@ -116,15 +116,15 @@ func notifySubscription(tid string, notification *notification.Notification) *sc
 		}
 
 		record := result.Records[0]
-		promise, err := record.Promise()
-		if err != nil {
+		promise, otherErr := record.Promise()
```
dfarr: how come this is otherErr?

internal/app/coroutines/rejectPromise.go
```diff
@@ -51,14 +51,14 @@ func RejectPromise(metadata *metadata.Metadata, req *t_api.Request, res func(*t_
...
-					Status: t_api.ResponseNotFound,
+					Status: t_api.StatusPromiseNotFound,
 				},
 			}, nil)
 		} else {
 			p, err := result.Records[0].Promise()
 			if err != nil {
 				slog.Error("failed to parse promise record", "record", result.Records[0], "err", err)
-				res(nil, err)
+				res(nil, t_api.NewResonateError(t_api.ErrAIOStoreSerializationFailure, "", err))
```
dfarr: should we put "failed to parse promise record" here?

internal/app/coroutines/rejectPromise.go
```diff
@@ -67,14 +67,14 @@ func RejectPromise(metadata *metadata.Metadata, req *t_api.Request, res func(*t_
 					c.Scheduler.Add(TimeoutPromise(metadata, p, RejectPromise(metadata, req, res), func(err error) {
 						if err != nil {
 							slog.Error("failed to timeout promise", "req", req, "err", err)
-							res(nil, err)
+							res(nil, t_api.NewResonateError(t_api.ErrAIOStoreFailure, "", err))
```
dfarr, line 70: here too

internal/app/coroutines/rejectPromise.go
dfarr: and for others with an empty string

guergabo: ah missed thanks !

internal/kernel/t_api/status.go
```diff
@@ -0,0 +1,119 @@
...
+// Platform level errors (5000-5999)
+
+type ResonateErrorCode int
+
+func (e ResonateErrorCode) String() string {
+	return strconv.Itoa(int(e))
+}
+
+const (
+	// catch call for now
```
```suggestion
	// catch all for now
```

internal/app/coroutines/notifiySubscriptions.go
guergabo: reverted, this was back when i had t_api.PlatformLevelErr defined above. it was a type issue, but all good now thanks.

internal/aio/aio.go
```diff
@@ -110,7 +111,7 @@ func (a *aio) Enqueue(sqe *bus.SQE[t_aio.Submission, t_aio.Completion]) {
 			slog.Debug("aio:enqueue", "sqe", sqe)
 			a.metrics.AioInFlight.WithLabelValues(sqe.Metadata.Tags.Split("aio")...).Inc()
 		default:
-			sqe.Callback(nil, fmt.Errorf("aio:subsystem:%s submission queue full", subsystem))
+			sqe.Callback(nil, t_api.NewResonateError(t_api.ErrAIOSubmissionQueueFull, "aio submission queue full", nil))
```
dfarr: let's put the subsystem name in the message

internal/api/api.go
```diff
@@ -104,8 +108,10 @@ func (a *api) Enqueue(sqe *bus.SQE[t_api.Request, t_api.Response]) {
 				status = int(res.CreateSubscription.Status)
 			case t_api.DeleteSubscription:
 				status = int(res.DeleteSubscription.Status)
-			default:
+			case t_api.Echo:
 				status = 200
```
dfarr: this should be 2000

dfarr (approved):

Issue #106, opened 2023-11-08
Add missing fields to proto spec


dfarr:
Missing fields:

- createdOn
- completedOn
- tags

dfarr:
Also need to include these values in the grpc responses

guergabo:
closed by above pr

PR #115, opened 2023-11-16
Fix: Add proto missing fields and add test for pkg/log

dreamjz:
Solve #106 
### Description
- Add missing fields for promise grpc response
- Add unit test for `pkg/log`


dfarr:
Thank you so much again @dreamjz !!! I wonder if we need to propagate these changes to the test harness in order for it to work.

@guergabo any inclination as to why the linearization test is failing?

guergabo:
99% percent sure its because i merged the test harness changes to adjust to the other PR before merging that PR. Let me get that merged in today and this should be good.

guergabo (approved):
Sorry @dreamjz for the blocker. Thanks for this amazing patch!

PR #144, opened 2023-12-08
Add ability to search promises by tag

dfarr:
Optionally search by tags, if tags are provided all key/value pairs must match exactly. Example
```json
{
  "a": "a",
  "b": "b"
}
```

Matches the promise:
```json
{
  "id": "foo",
  "tags": {"a": "a", "b": "b", "c": "c"}
}
```

But not the promise:
```json
{
  "id": "bar",
  "tags": {"a": "a"}
}
```

Additionally, we extract the tag `resonate:invocation` into a generated column and add an index on this column. This tag/column will be used to search for "top level" promises. This detail is not part of the promise specification and if not implemented will fall back to a standard tag search.


internal/app/subsystems/api/service/service_test.go
```diff
@@ -1,653 +0,0 @@
-package service
```
dfarr, line 1: Removed because these are duplicates of the http/grpc tests

PR #160, opened 2023-12-19
Allow slashes in promise name

dfarr:
Now that promise id constitutes the final part of the http route we can allow slashes in the id. Gin allows this with special character `*` but we need to strip the leading `/` which gin will preserve.


guergabo (approved):
lgtm

PR #161, opened 2023-12-19
Refactor schedule promises

dfarr:
Remove crontime and use `schedule.NextRunTime` directly, perhaps lower cognitive overhead with one less variable.


guergabo (approved):
lgtm

PR #167, opened 2023-12-20
Refactor commands

dfarr:
Refactors the serve and dst commands to conform to new style.

Additionally:
- changes the client to use a response client
- minor changes to the openapi spec
- pretty prints promises and schedule responses
- adds tick time to promise timeout time in schedule promises coroutine
- some additional renaming


api/promises-openapi.yml
```diff
@@ -53,7 +53,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/PromiseList"
+                $ref: "#/components/schemas/PromiseSearchResponse"
```
guergabo, line 56: let's update durable-promise repo with this new spec.

internal/app/coroutines/schedulePromises.go
```diff
@@ -108,7 +108,7 @@ func schedulePromise(tid string, schedule *schedule.Schedule) *scheduler.Corouti
 								Id:             id,
 								State:          state,
 								Param:          schedule.PromiseParam,
-								Timeout:        schedule.PromiseTimeout,
+								Timeout:        crontime + schedule.PromiseTimeout,
```
guergabo, line 111: nice catch.

cmd/promises/search.go
```diff
@@ -0,0 +1,92 @@
...
+					promise, err := json.Marshal(p)
+					if err != nil {
+						cmd.PrintErr(err)
+						continue
+					}
+
+					cmd.Println(string(promise))
+				}
+			} else {
+				prettyPrintPromises(cmd, *resp.JSON200.Promises...)
```
guergabo, line 81: nit: smart formatting thing. let's do if { return } prett instead of if else{} more idiomatic and nicer to read. would love this standard for this repo in general.

guergabo (approved):

PR #168, opened 2023-12-21
Add ability to search by tags on cli

dfarr:


api/promises-openapi.yml
```diff
@@ -151,7 +151,14 @@ components:
               - "abc123*" - Would match ID "abc123def" 
               - "pro*fil" - Would match IDs containing "pro123file"
           state:
-            $ref: "#/components/schemas/PromiseState"
+            type: string
+            enum:
+              - pending
```
guergabo, line 156: same thing lets update durable-promise repo and think of long term plan to synchronize. i think this should be the source of truth and repo can sync on every merge? let's think abotu this

guergabo (approved):
lgtm

PR #175, opened 2023-12-28
add grpc support for schedules api

guergabo:
Changes: 
- grpc support for schedules api 
- include headers for schedules  
- fix client api issue (stuck it in there) 
- re-enable test harness
- remove inline in openapi for resuability and removing anonymous structs (search queries are find since just strings and single usage)


api/promises-openapi.yml
```diff
@@ -266,3 +259,13 @@ components:
             type: string
         data:
           type: string
+
+    SearchPromisesResponseObj:
```
dfarr: Can we leave this inline?

internal/app/subsystems/api/grpc/api/schedule.proto
```diff
@@ -0,0 +1,81 @@
...
+    int64 lastRunTime = 9;
+    int64 nextRunTime = 10; 
+    string idempotencyKey = 11;
+    int64 createdOn = 12;    
+}
+
+// CREATE 
+
+message CreateScheduleRequest {
+    Schedule schedule = 1;
```
dfarr: I think we should manually list the fields that can be specified here since it is only a subset

internal/app/subsystems/api/grpc/grpc.go
```diff
@@ -33,6 +33,7 @@ func New(api api.API, config *Config) api.Subsystem {
 
 	server := grpc.NewServer(grpc.UnaryInterceptor(s.log)) // nosemgrep
 	grpcApi.RegisterPromiseServiceServer(server, s)
```
dfarr: Not required for this PR, but I wonder how the name became `PromiseServiceServer`? Would be nice to align on:
- PromiseServer
- SchedulesServer

internal/app/subsystems/api/grpc/grpc_test.go
```diff
@@ -23,10 +24,11 @@ import (
...
 	*test.API
-	subsystem api.Subsystem
-	errors    chan error
-	conn      *grpc.ClientConn
-	client    grpcApi.PromiseServiceClient
+	subsystem      api.Subsystem
+	errors         chan error
+	conn           *grpc.ClientConn
+	client         grpcApi.PromiseServiceClient
+	scheduleClient grpcApi.SchedulesClient
```
dfarr, line 31: At first, I liked that the promise and schedule clients were separate. But the tests made me realize that the user would also need to instantiate a different client for schedules :(

Do you think think we should combine into one client? Maybe there is a `protoc` command we could use to combine multiple files into a single client and that way we can still keep the files separated (which would be nice for syncing with the durable promises specification repo).

internal/app/subsystems/api/grpc/grpc_test.go
```diff
@@ -925,3 +930,298 @@ func TestRejectPromise(t *testing.T) {
...
+	}
+
+	tcs := []struct {
+		name    string
+		grpcReq *grpcApi.SearchSchedulesRequest
+		req     *t_api.Request
+		res     *t_api.Response
+	}{
+		{
+			name: "SearchPromises",
```
dfarr:
```suggestion
			name: "SearchSchedules",
```

api/promises-openapi.yml
guergabo: this is an openapi thing. u shouldnt do inline in general cause it results in anonymous structs making it harder to deserialize.

dfarr: Then we need to be consistent and change everywhere, the commands (that use the generated code) will need to be updated and we should change search schedules definition to match.

internal/app/subsystems/api/grpc/api/schedule.proto
guergabo: this is not the only place we use schedule type. that would just give us a lot of boilerplate (copy paste) every time the type schedule is used. Plus, for protos it's better to do this since we can change the schedule schema without breaking as much as if it was inline. 

also this is consistent with the promise proto which is not inline assuming for the same reason.

api/promises-openapi.yml
guergabo: yea np. in general let's avoid inline stuff. in my experience, it causes trouble down the line with different tooling. def get the reasoning though (so you dont jump back and forth)

internal/app/subsystems/api/grpc/grpc.go
guergabo: it became promiseservice because that is the name of the service "PromiseService" in the proto. I can just change it to service "Promise" and will be the same.

internal/app/subsystems/api/grpc/grpc_test.go
guergabo: this is exactly how we do the cli currently. since we have one ClientSet with multiple clients. 

i'm not sure the context a user instantiate this client directly. but from the sdk we can do a clientSet with clients like cli. 

in general i like clientset style. keeps things separate and clean.

internal/app/subsystems/api/grpc/grpc.go
guergabo: it's PromiseService because the type promise is already defined in proto.

internal/app/subsystems/api/grpc/api/schedule.proto
dfarr: This is not consistent with promises, see [CreatePromiseRequest](https://github.com/resonatehq/resonate/blob/main/internal/app/subsystems/api/grpc/api/promise.proto#L63-L70) for example, we have a subset of the Promise object.

guergabo: that's openapi. look at current promise.proto

guergabo: oh i see what ur saying now. was misundertanding before. let's chat about this. after if we want the experience of grpc to be consistent with http experieence we should prob update promise proto. we do similar with status. let's chat.

api/promises-openapi.yml
```diff
@@ -266,3 +242,41 @@ components:
             type: string
         data:
           type: string
+
+    CreatePromiseReq:
+      type: object
+      reuired:
```
dfarr:
```suggestion
      required:
```

api/schedules-openapi.yml
```diff
@@ -90,14 +79,7 @@ paths:
...
               schema:
-                type: object
-                properties:
-                  cursor:
-                    type: string
-                  schedules:
-                    type: array
-                    items:
-                      $ref: "#/components/schemas/Schedule"
+                $ref: "#/components/schemas/SearchSchedulesResponseObj"
```
dfarr:
```suggestion
                $ref: "#/components/schemas/SearchSchedulesResp"
```

api/schedules-openapi.yml
```diff
@@ -200,3 +198,13 @@ components:
           type: object
           additionalProperties:
             type: string
+
+    SearchSchedulesResponseObj:
```
dfarr:
```suggestion
    SearchSchedulesResp:
```

cmd/promises/create.go
```diff
@@ -48,8 +49,8 @@ func CreatePromiseCmd(c client.ResonateClient) *cobra.Command {
 			}
 
 			body := promises.CreatePromiseJSONRequestBody{
-				Id:      id,
```
dfarr: Hmm this wasn't a pointer before, maybe it was the typo?

internal/app/subsystems/api/grpc/grpc.go
dfarr: Howcome that's not the case for schedule as well? We have a "Schedule" in the schemas there too

internal/app/subsystems/api/grpc/grpc_test.go
```diff
@@ -46,11 +48,12 @@ func setup() (*grpcTest, error) {
...
-		API:       api,
-		subsystem: subsystem,
-		errors:    errors,
-		conn:      conn,
-		client:    grpcApi.NewPromiseServiceClient(conn),
+		API:            api,
+		subsystem:      subsystem,
+		errors:         errors,
+		conn:           conn,
+		client:         grpcApi.NewPromiseServiceClient(conn),
```
dfarr: can we rename this to promiseClient?

guergabo: i commented above why we can't

guergabo: unless we change promise type

dfarr (approved):

PR #176, opened 2023-12-29
Contributing to the documentation - Architecture

zeu5:
Hello @guergabo!

I've taken an ambitious step of understanding and updating the documentation of Resonate codebase. Based on my understanding I've updated the following, 

Changes:
- Added information about the kernel subsystem in `docs/architecture.md`
- A new figure describing the lifecycle of a coroutine in resonate

While this is a small step, I wish to use this as an opportunity to understand the requirements by the maintainers for the documentation. Hoping to improve with feedback!


docs/architecture.md
```diff
@@ -10,23 +10,94 @@ Resonate consists of three subsystems:
...
+}
+
+type CQE[I Input, O Output] struct {
+    *MetaData
+    Callback func(...)
+    ...
+}
+```
+
+For every `SQE` within the Kernel, there exists a corresponding `CQE` and the callback and metadata is shared between the two instances.
```
guergabo, line 45: very nice insight here.

guergabo (approved):

PR #187, opened 2024-01-05
Upload artifacts directly to release

dfarr:
Releases can be created through the standard release interface.

Example: 
- https://github.com/dfarr/resonate/actions/runs/7426032236
- https://github.com/dfarr/resonate/releases/tag/test-3


guergabo (approved):
lgtm

PR #189, opened 2024-01-08
test(dst): add support for simulating cron schedules

guergabo:
I played around with many more "invasive" approaches and backed off a bit after some trial and error and landed to this solution. I like this solution because it is simple yet effective for our use case. 

### Changes
- **_logical clocks_**: extend the util.next() to support **_logical clocks_** not just physical clocks based on the cron schedule. Effectively eliminating real time from the system by translating minutes to ticks. 
- **_client-centric model_**: stay within the **_client-centric model_** of dst to test if schedules are being executed on time (a.k.a bulk validate within searchSchedules). 
- **_current tick_**: modify the mode.step() to capture **_current tick_** information. this helps with verification of correctness. 
- **_inflight cache_**: optimize scheduling variance by using an **_inflight cache_** (similar to notifications). there was a noticeable performance difference from < 500 to < 100 ticks variance.

### DIFF
--- a/internal/kernel/t_api/response.go
+++ b/internal/kernel/t_api/response.go
@@ -67,7 +67,7 @@ type CreateScheduleResponse struct {
 type SearchSchedulesResponse struct {
 	Status    ResponseStatus                  `json:"status"`
 	Cursor    *Cursor[SearchSchedulesRequest] `json:"cursor,omitempty"`
-	Schedules []*schedule.Schedule            `json:"promises,omitempty"`
+	Schedules []*schedule.Schedule            `json:"schedules,omitempty"`
 }
 
 type ReadScheduleResponse struct {
--- a/test/dst/dst.go
+++ b/test/dst/dst.go
@@ -123,7 +123,7 @@ func (d *DST) Run(r *rand.Rand, api api.API, aio aio.AIO, system *system.System,
 			i++
 		}
 
-		system.Tick(t, nil)
+		system.Tick(t*50_000, nil)
 
 		if len(errs) > 0 {
 			break
--- a/test/dst/dst.go
+++ b/test/dst/dst.go
@@ -15,15 +15,16 @@ import (
 )
 
 type Config struct {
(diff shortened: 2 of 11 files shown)


internal/kernel/t_api/response.go
```diff
@@ -67,7 +67,7 @@ type CreateScheduleResponse struct {
 type SearchSchedulesResponse struct {
 	Status    ResponseStatus                  `json:"status"`
 	Cursor    *Cursor[SearchSchedulesRequest] `json:"cursor,omitempty"`
-	Schedules []*schedule.Schedule            `json:"promises,omitempty"`
+	Schedules []*schedule.Schedule            `json:"schedules,omitempty"`
```
dfarr, line 70: Nice catch!

test/dst/generator.go
```diff
@@ -292,11 +294,11 @@ func (g *Generator) GenerateSearchSchedules(r *rand.Rand, t int64) *t_api.Reques
 
 func (g *Generator) GenerateCreateSchedule(r *rand.Rand, t int64) *t_api.Request {
 	id := g.idSet[r.Intn(len(g.idSet))]
-	cron := fmt.Sprintf("%d %d * * *", r.Intn(60), r.Intn(24))
+	cron := fmt.Sprintf("%d * * * *", RangeIntn(r, 10, 60))
```
dfarr: Could we go from [0, 60)?

test/dst/dst.go
```diff
@@ -111,7 +112,7 @@ func (d *DST) Run(r *rand.Rand, api api.API, aio aio.AIO, system *system.System,
 				Metadata:   metadata,
 				Submission: req,
 				Callback: func(res *t_api.Response, err error) {
-					modelErr := model.Step(req, res, err)
+					modelErr := model.Step(reqTime, req, res, err)
```
dfarr: Can we remove reqTime from the step function? I think we ended up not using it

Issue #97, opened 2023-11-04
Create fresh db every time


guergabo:
Currently, if we run `resonate serve` it creates a `resonate.db` file. Every time I rerun the server it DOES NOT create a new `resonate.db`, but rather uses the old one implicitly. 

It is not explicit that, by default, the db lives beyond the lifecycle of the server. This could lead to confusion and end users getting unexpected test results. 

We want a fresh db every time the server is started. Let's either delete the db, and/or startup the server with a new filename.

dfarr:
For durability we need the database to survive application restarts, this should be the default behaviour. We can provide an optional flag to reset the db on shutdown, there already is a lifecycle hook in the code called `Reset` that implements the logic to do this.

pancholinaman07:
For fresh db every time we can just add the query to delete database if already exist

favalos:
Looking forward to do this change.

Should we add a flag for SQLLite to Reset?

BTW, can I get the link to join Slack?

guergabo:
@favalos yea i think a flag would be a nice approach. i'll try to get to this soon, but feel free to submit a PR. as @dfarr mentioned, there is  Reset() function that already implements this logic, just need to trigger it conditionally based on whether or not the flag has been set. 

slack is on the readme, but here you go: https://resonatehqcommunity.slack.com/

PR #202, opened 2024-01-10
Add Flag to Reset the DB on Shutdown.

favalos:
Pull request to add the flag discussed in #97.

I had to add a reset flag to every DB configuration and change the Reset implementation.

Other approach could be doing the implementation on the Shutdown method, but that would imply to add a new `Resetable() bool` method to the subsystem interface.

Please let me know your comments.


dfarr:
Thank you so much for the contribution @favalos!

I have one (hopefully) small change to ask.

The shutdown function on the api/aio interfaces are used as a signal to _begin_ a graceful shutdown. The aio method is currently unused, but we do use it on the api side. When this signal is received we stop taking new requests but continue to process any sqes/cqes already in the system until all are complete. As such we can't reset the database via the shutdown function as we still need to write to the database after this function has been called.

You can see the shutdown process [here](https://github.com/resonatehq/resonate/blob/main/cmd/serve/serve.go#L148-L170).

1. shutdown signal received (happens in a go routine and calls api.shutdown+aio.shutdown via `system.Shutdown`)
2. `system.Loop` will return once all outstanding sqes/cqes are complete
3. `api.Stop()` called
4. `aio.Stop()` called (internally calls Stop for all subsystems)

I recommend moving the call to Reset (if the flag is true) the Stop function, this is when we can safely reset.

favalos:
Hi @dfarr,

Thanks for the detailed explanation, I did the changes accordingly.

Let me know if that looks fine.

guergabo:
Thank you @favalos for contributing to resonate !!!

dfarr (approved):
Looks great! Thank you @favalos

dfarr:
Closes #97

Issue #166, opened 2023-12-20
Collapse the Resolve, Reject, Cancel Coroutines to 1 Complete Coroutine


guergabo:
It really only is 1 type of operation. same logic.

favalos:
Can you elaborate more on this? No difference in the response?

guergabo:
Hi @favalos, sorry for the ambiguity. What I meant is the following: 

1. Recently, we did a refactor to our promise API and the resolve, reject, and cancel promise operations are all behind a PATCH (http) or COMPLETE (grpc) api endpoint. Look here: https://github.com/resonatehq/resonate/blob/main/internal/app/subsystems/api/http/promise.go#L105
2. If you see we are switching and choosing one of the three operations, but if you follow down the path of "completing a promise", the business logic is exactly the same minus the actual "state transition" (pending to resolve, reject, or cancel).
3. Therefore, we can get rid of the copy paste resolve, reject and cancel coroutines and just have one "completePromise" coroutine that handles any state transition from pending to "resolve, reject, or cancel" to limit the boilerplate and similar logic.

Hope this makes sense. If you are interested, this would be a great issue to get an understanding of our system. LMK if you have any follow-up questions.

favalos:
Hey @guergabo,

I am working on this, I am very interested in learning more about resonate. Will let you know if I have questions.

Thanks!

guergabo:
Awesome to hear, @favalos ! Looking forward to your questions and PR 🔥

favalos:
Thanks!

dfarr:
Thank _you_ @favalos !!!

PR #218, opened 2024-01-25
Changes for Complete coroutine.

favalos:
WIP for #166.

@guergabo I did the changes for this issue.

I marked this as Draft because I am getting a couple of test failures, if you have any advice it would be great. 

Any comment or suggestions are welcome.

Thanks!


dfarr:
Thank you for the PR again @favalos!

I think the reason you are seeing test failures is because you re-ordered the states [here](https://github.com/resonatehq/resonate/pull/218/files#diff-46480e4c0b744a50df1e74167003c3579eda8ee719e1a47b7c31dcc1ba69c905R46-R47). So now the constants map to:
```go
const (
	Pending State = 1 << iota
	Resolved // 2
	Rejected // 4
	Canceled // 8
	Timedout // 16
	Invalid  // 32
)
```

I like this ordering better but we will need to change the test cases to match for the `Timedout` case.

We will also need to remove Invalid from this enum and guard against ever getting to this point. Instead I recommend changing the type of State in the `CompletePromiseBody` to `promise.State` that way we can take advantage of the [UnmarshalJSON function](https://github.com/resonatehq/resonate/blob/0773faba49a50a2ce85d1e8b0029909f8783f364/pkg/promise/promise.go#L70) we already have and the validation would become

```
!body.State.In(promise.Resolved|promise.Rejected|promise.Canceled)
```

favalos:
Hey @dfarr,

Thanks for the hint, I re-ordered the Enum and that did the trick.

I fixed another couple of errors that I found and did the changes related to the recommendation mentioned above.

Thank you!

dfarr:
Looks great @favalos! The only other change I would recommend would be consolidating the t_api Kind to a single kind for these requests called `CompletePromise`. This is a lot of copypasta work so let me know if you want me to do that and I can add a quick commit.

favalos:
I can work on that tonight, no worries! 

Thanks!

favalos:
@dfarr

Done! Let me know any other suggestions.

Thanks!

internal/app/subsystems/api/http/promise.go
```diff
@@ -122,17 +121,11 @@ func (s *server) completePromise(c *gin.Context) {
...
-	switch strings.ToUpper(body.State) {
-	case promise.Resolved.String():
-		resp, err = s.service.ResolvePromise(id, &header, body)
-	case promise.Rejected.String():
-		resp, err = s.service.RejectPromise(id, &header, body)
-	case promise.Canceled.String():
-		resp, err = s.service.CancelPromise(id, &header, body)
-	default:
+	if !body.State.In(promise.Resolved | promise.Rejected | promise.Canceled) {
 		c.JSON(http.StatusBadRequest, api.HandleValidationError(errors.New("invalid state")))
```
dfarr, line 125: We just need a return after this line, sorry I didn't see that earlier

dfarr:
Awesome thank you so much @favalos 🎉🎉 🎉 

We just need one return statement and I will merge this PR :)

internal/app/subsystems/api/http/promise.go
favalos: I added the return after this line, thanks!

PR #244, opened 2024-03-13
quickstart cmd for cli

vaibhawvipul:
- adding quickstart cmd

### DIFF
--- a/cmd/quickstart/quickstart.go
+++ b/cmd/quickstart/quickstart.go
@@ -1,0 +1,32 @@
+package quickstart
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func NewCmd() *cobra.Command {
+	var quickstartcmd = cobra.Command{
+		Use:   "quickstart [sdk]",
+		Short: "Create a minimal getting started app for the specified SDK",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			sdk := args[0]
+			switch strings.ToLower(sdk) {
+			case "ts", "typescript":
+				printInstructionsForTypeScript()
+			default:
+				fmt.Printf("Error: Unknown SDK '%s'.\n", sdk)
+			}
+		},
+	}
+	return &quickstartcmd
+}
+
+func printInstructionsForTypeScript() {
+	fmt.Println("Getting started with Resonate for TypeScript")
+	fmt.Println("> 🏴‍☠️ 1. Run git clone git@github.com:resonatehq/ts-quickstart.git quickstart && cd quickstart")
+	fmt.Println("> 🏴‍☠️ 2. Follow the Getting Started instructions in Readme.md")
+}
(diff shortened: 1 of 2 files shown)


cmd/quickstart/quickstart.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/spf13/cobra"
+)
+
+func NewCmd() *cobra.Command {
+	var quickstartcmd = cobra.Command{
+		Use:   "quickstart [sdk]",
+		Short: "Create a minimal getting started app for the specified SDK",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			sdk := args[0]
```
guergabo, line 13: lets just add a check in case they don't add any arg so we don't panic.

cmd/quickstart/quickstart.go
```diff
@@ -0,0 +1,32 @@
...
+				fmt.Printf("Error: Unknown SDK '%s'.\n", sdk)
+			}
+		},
+	}
+	return &quickstartcmd
+}
+
+func printInstructionsForTypeScript() {
+	fmt.Println("Getting started with Resonate for TypeScript")
+	fmt.Println("> 🏴‍☠️ 1. Run git clone git@github.com:resonatehq/ts-quickstart.git quickstart && cd quickstart")
```
guergabo: nit: just pass in "ts" and "TypeScript" so this function can be used for any language/sdk.

`printInstructionsFor(language, alias string)`

cmd/quickstart/quickstart.go
```diff
@@ -0,0 +1,32 @@
...
+		Use:   "quickstart [sdk]",
+		Short: "Create a minimal getting started app for the specified SDK",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			sdk := args[0]
+			switch strings.ToLower(sdk) {
+			case "ts", "typescript":
+				printInstructionsForTypeScript()
+			default:
+				fmt.Printf("Error: Unknown SDK '%s'.\n", sdk)
```
guergabo: also look at the complete promise cmd for example. 

use cmd.* for any io and distinguish between printf and printerr. this should be err. 

we do this to follow dependency injection principle.

guergabo (changes requested):

cmd/quickstart/quickstart.go
guergabo: oh wait i see ur using `Args:  cobra.ExactArgs(1)` does that fail if there isn't one ? I haven't used that feature?

guergabo (approved):

cmd/quickstart/quickstart.go
vaibhawvipul: Yes, you're correct. The Args: cobra.ExactArgs(1) configuration in Cobra ensures that the command expects exactly one argument. If the user doesn't provide exactly one argument when executing the command, Cobra will display an error message and the command won't be executed.

guergabo: I like it can u create an issue to track doing this for all the other commands for consistency?

guergabo: if you wanna just do it feel free too.

vaibhawvipul:
works as expected

guergabo (approved):
lgtm

vaibhawvipul:
works as expected 
Screenshot 2024-03-13 at 9 52 34 PM

PR #263, opened 2024-03-25
feat(ms): translate to ms

### LINKED ISSUE #217: Consolidate time specification to one of ms or s
Promises:
- promise timeout time is specified in absolute ms
- created_at, completed_at is in ms

Schedules
- promise timeout time is specified in absolute ms
- created_at is in ms

Locks
- lock expiry_in_seconds is specified in relative s
- expires_at in relative ms

I propose we choose either seconds or milliseconds and stick to it consistently, aka what granularity do we want to expose to the user? On a database level let's store everything as ms so as to maintain maximum granularity if we ever choose to switch.

Corollary:  should we indicate the units in the key name, eg: `timeout` or `timeoutInMs`? Whichever we choose we should also use consistently.

guergabo:
Fixes https://github.com/resonatehq/resonate/issues/217

Separate PR updates the SDK.


internal/app/subsystems/api/grpc/lock.go
```diff
@@ -30,15 +30,15 @@ func (s *server) AcquireLock(ctx context.Context, req *grpcApi.AcquireLockReques
 	if req.Lock.ExecutionId == "" {
 		return nil, grpcStatus.Error(codes.InvalidArgument, "lock.execution_id must be provided")
 	}
-	if req.Lock.ExpiryInSeconds == 0 {
+	if req.Lock.ExpiryInMilliseconds == 0 {
 		return nil, grpcStatus.Error(codes.InvalidArgument, "lock.timeout must be provided")
```
dfarr: should this warning by lock.ExpiryInMilliseconds?

dfarr (approved):
Nice!

Issue #266, opened 2024-03-26
Assertion on SQE/CQE Callback


dtornow:
# Problem

During our recent experimentation with the DST, it became apparent that our system is susceptible to invariant violations which currently go undetected

I believe there are two invariants invariants in the system:

1. **SQE** At the point of enqueueing, every Submission Queue Entry (SQE) must have an associated callback set.
2. **CQE** At the point of dequeuing, every Completion Queue Entry (CQE) must have an associated callback set.

# Proposed Solution

To address this issue, I propose the integration of two assertions within the scheduler

1. For the SQE Invariant: Insert an assertion to verify that a callback is set for each SQE before it is enqueued.
2. For the CQE Invariant: Insert an assertion to ensure that a callback is set for each CQE after it is dequeued.

# Conclusion and Call to Action

Ensuring our system's invariants is crucial for maintaining correctness. By encoding these invariants as assertions, we take a proactive step towards ensuring correctness. This practice should be extended to encompass more system invariants, further embedding the principle of fail-fast and ensuring that any violations are promptly identified and addressed.

I invite feedback on this proposal and encourage a discussion on identifying additional invariants that could benefit from similar enforcement mechanisms.

guergabo:
it appears as if the sqe callback is already enforced and set by the scheduler: https://github.com/resonatehq/resonate/blob/4c1b89c625828ba72787c0abe84c55bd77d26a98/internal/kernel/scheduler/scheduler.go#L52 

We can still assert the cqe callback a few lines above.

guergabo:
we can also do something like this where we set it before calling enqueue:  

`
sqe := &bus.SQE[t_aio.Submission, t_aio.Completion]{}
util.Assert(sqe.callback != nil)
s.aio.Enqueue(sqe)
`

aneessh18:
> we can also do something like this where we set it before calling enqueue:
> 
> `sqe := &bus.SQE[t_aio.Submission, t_aio.Completion]{} util.Assert(sqe.callback != nil) s.aio.Enqueue(sqe)`

can we close this issue ?

PR #267, opened 2024-03-27
DST fault injection bug fix

vaibhawvipul:
- handle aio errors
- add assertions (fixes - [issue 266](https://github.com/resonatehq/resonate/issues/266))


internal/aio/aio_dst.go
```diff
@@ -95,27 +94,27 @@ func (a *aioDST) Flush(t int64) {
...
+				}
+				a.cqes = append(a.cqes, processedCQEs...)
+			} else {
+				// error without processing
+				res := make([]*bus.CQE[t_aio.Submission, t_aio.Completion], len(sqes.Value))
+				for i := range sqes.Value {
+					res[i] = &bus.CQE[t_aio.Submission, t_aio.Completion]{
+						Metadata:   sqes.Value[i].Metadata,
+						Completion: nil,
+						Callback:   sqes.Value[i].Callback,
```
guergabo, line 112: excellent here

internal/aio/aio_dst.go
```diff
@@ -95,27 +94,27 @@ func (a *aioDST) Flush(t int64) {
...
-					for _, cqe := range processedCQEs {
-						errorCqe := createErrorCQE("aio dst error occurred after processing SQE", "aio dst failure", simpleCallback)
-						errorCqe.Completion = &t_aio.Completion{}
-
-						slog.Debug("aio dst: failure", "err", errorCqe, "cqe", cqe)
-						a.cqes = append(a.cqes, cqe, errorCqe)
+				for i := range processedCQEs {
+					if a.r.Float64() < a.failureProbability && processedCQEs[i].Error == nil {
+						processedCQEs[i].Error = fmt.Errorf("aio dst: failure, after processing")
+					}
```
guergabo: can we be more literal "error: failure injection after processing" + "error: failure injection before processing" ?

test/dst/model.go
```diff
@@ -143,6 +143,10 @@ func (m *Model) Step(t int64, req *t_api.Request, res *t_api.Response, err error
 			return nil
 		case t_api.ErrAIOSubmissionQueueFull:
 			return nil
+		case t_api.ErrAIOStoreFailure:
+			if strings.Contains(resErr.Unwrap().Error(), "aio dst: failure") {
```
guergabo: don't do strings.Contains. do errors.Is or errors.As if you can. unwrap() returns the actual error type.

guergabo: .unwrap() no .Error()

test/dst/dst_test.go
```diff
@@ -146,7 +147,16 @@ func TestDST(t *testing.T) {
 	})
 
 	if errs := dst.Run(r, api, aio, system, reqs); len(errs) > 0 {
-		t.Fatal(errs)
+		// if there are errors which have aio dst in the message, then it's not a real error
```
guergabo: we don't need this. just have model.step take care of it.

test/dst/dst.go
```diff
@@ -122,16 +122,25 @@ func (d *DST) Run(r *rand.Rand, api api.API, aio aio.AIO, system *system.System,
...
+			}
+
 			api.Enqueue(&bus.SQE[t_api.Request, t_api.Response]{
 				Metadata:   metadata,
 				Submission: req,
 				Callback: func(res *t_api.Response, err error) {
-					modelErr := model.Step(t, req, res, err)
-					if modelErr != nil {
-						errs = append(errs, modelErr)
+					// CQE Invariant: Assert that a callback is set after dequeuing
```
guergabo: for asserts we use util.Assert not this. and the invariant shouldn't be asserted here but in the schedule for production.

guergabo (changes requested):

internal/aio/aio_dst.go
guergabo: remove `processedCQEs[i].Error == nil`

guergabo: var ErrFailureInjection  = errrors.New("")

guergabo:
fyi: we had pair programming session this morning and few additional things were discovered / discussed. In particular the handling of failure injection post-process. Since the model will encounter reads that it was not aware of since the client believed the AIOStoreFailure occurred. Good stuff!

internal/aio/aio_dst.go
```diff
@@ -95,27 +95,25 @@ func (a *aioDST) Flush(t int64) {
...
+				a.cqes = append(a.cqes, processedCQEs...)
+			} else {
+				// error without processing
+				res := make([]*bus.CQE[t_aio.Submission, t_aio.Completion], len(sqes.Value))
+				for i := range sqes.Value {
+					res[i] = &bus.CQE[t_aio.Submission, t_aio.Completion]{
+						Metadata:   sqes.Value[i].Metadata,
+						Completion: nil,
+						Callback:   sqes.Value[i].Callback,
+						Error:      errors.New("aio dst: failure, before processing"),
```
guergabo: can we make this and the other error variables above the file: 

ErrFailureInjectionBeforeAIOStoreProcessing = errors.New("")
ErrFailureInjectionAfterAIOStoreProcessing = errors.New("")

internal/kernel/scheduler/scheduler.go
```diff
@@ -54,6 +55,7 @@ func (s *Scheduler) Tick(t int64, batchSize int) {
 
 	// dequeue cqes
 	for _, cqe := range s.aio.Dequeue(batchSize) {
+		util.Assert(cqe.Callback != nil, "cqe is nil")
```
guergabo, line 58: can we change the assert message to "cqe must not be nil" ? that way the message matches the condition and don't have to think backwards when reading it as documentation.

internal/kernel/scheduler/scheduler.go
```diff
@@ -88,7 +90,11 @@ func (s *Scheduler) Tick(t int64, batchSize int) {
 						}
 					}
 				},
-			})
+			}
+
+			util.Assert(sqe.Callback != nil, "sqe is nil")
```
guergabo: can we add two additional assertions: one making sure metadata is not nil and same for submission?

test/dst/model.go
```diff
@@ -143,6 +143,8 @@ func (m *Model) Step(t int64, req *t_api.Request, res *t_api.Response, err error
 			return nil
 		case t_api.ErrAIOSubmissionQueueFull:
 			return nil
+		case t_api.ErrAIOStoreFailure:
+			return nil
```
guergabo: errors.Is here to differentaite from expected to unexpected. tolerated vs untolated.  did we inject this error or did it happen on its own? is what we want to answer. 

errors.is not sttring so that it doesn't change eveen if we change the string value of the error.

guergabo (changes requested):

internal/kernel/scheduler/scheduler.go
guergabo: also can we assert the other fields of cqe are not nil either. that way we are certain the loop was correct.

internal/kernel/scheduler/scheduler.go
dfarr: I'm not sure that we need this assertion because we define callback inline so it's guaranteed to be set. We could assert `submission` and `coroutine.metadata` as Gabriel suggested before constructing the sqe

test/dst/dst.go
```diff
@@ -126,11 +127,10 @@ func (d *DST) Run(r *rand.Rand, api api.API, aio aio.AIO, system *system.System,
 				Metadata:   metadata,
 				Submission: req,
 				Callback: func(res *t_api.Response, err error) {
-					modelErr := model.Step(t, req, res, err)
+					modelErr := model.Step(t, req, res, err, d.config.FaultInjectionMode)
```
dfarr: Rather than pass this in for every step, let's instantiate the model with the "scenario" information since this does not change. I think we could do that [here](https://github.com/resonatehq/resonate/blob/main/test/dst/dst.go#L45).

test/dst/model.go
dfarr: I also think we should do this cross-cutting where we inspect the error type to see if it an injected failure regardless of the code.

test/dst/model.go
```diff
@@ -59,7 +59,10 @@ type Schedules map[string]*ScheduleModel
 type Subscriptions map[string]*SubscriptionModel
 type Locks map[string]*LockModel
 type Tasks map[string]*TaskModel
-type ResponseValidator func(int64, *t_api.Request, *t_api.Response) error
+type ResponseValidator func(int64, *t_api.Request, *t_api.Response, bool) error
+
+// create a not found error for the given kind assign that to a var
+var ErrNotFound = errors.New("not found")
```
dfarr: Can we remove this?

cmd/dst/run.go
```diff
@@ -83,9 +84,15 @@ func RunDSTCmd() *cobra.Command {
 
 			go metricsServer.ListenAndServe() // nolint: errcheck
 
+			scenario := dst.Default
+			aioScenario := aio.Default
```
dfarr: Let's just have one scenario on the DST rather than duplicate this in the aio

cmd/dst/run.go
```diff
@@ -83,9 +84,15 @@ func RunDSTCmd() *cobra.Command {
...
+			if faultInjectionMode > 0 {
```
dfarr: Can we have the user pass in the scenario rather than an integer?

cmd/dst/run.go
```diff
@@ -171,7 +178,7 @@ func RunDSTCmd() *cobra.Command {
 				Tags:            tags.Resolve(r),
 				Urls:            urls.Resolve(r),
 				Retries:         retries.Resolve(r),
-			})
+			}, scenario)
```
dfarr: Let's just put this in the config, similar to all the configs above

internal/aio/aio_dst.go
```diff
@@ -1,33 +1,41 @@
...
 
+type Scenario int
 type aioDST struct {
 	r                  *rand.Rand
 	sqes               []*bus.SQE[t_aio.Submission, t_aio.Completion]
 	cqes               []*bus.CQE[t_aio.Submission, t_aio.Completion]
 	subsystems         map[t_aio.Kind]Subsystem
 	metrics            *metrics.Metrics
 	failureProbability float64
+	faultInjectionMode Scenario
```
dfarr: Let's just call this scenario

PR #269, opened 2024-04-02
[WIP] Announcements and Monitoring of DST

### LINKED ISSUE #268: Add Announcements

Resonate's Deterministic Simulation Testing cannot observe the Resonate Server beyond the request/response API. E.g. the Deterministic Simulation Testing cannot observe if notifications are actually sent.

[P-Lang](https://github.com/p-org/P) style announcements & monitors. With announcements and monitors, DST will be able to verify that e.g. announcements are sent if a corresponding subscription exists.

## Implementation Details

**Step 1**

Create a package `announcements`.

**Step 2**

Define an interface `Announcement` with a function

```golang
type Announcement interface {
    Announce(data map[string]string)
}
```

**Step 3**

Define two different implementations:
- `NopAnnouncement`
- `DstAnnouncement`

**Step 4**

Use a package-level private variable to hold a singleton instance. Instantiate the correct singleton instance based on config parameter on startup e.g.

```golang
package announcements

import "sync"

var (
    instance Announcement
    once     sync.Once
)

func Initialize(envType EnvironmentType) {
    once.Do(func() {
        switch envType {
        case Nop:
            instance = &NopAnnouncement{}
        case Dst:
            instance = &DstAnnouncement{}
        default:
            // Handle default case or throw an error
        }
    })
}

// Additional code to define EnvironmentType, NopAnnouncement, and DstAnnouncement follows...

```

**Step 5**

Implement the interface

- `Nop` This will simply implement the Announce method without performing any operations.
- `Dst` This will need to safely add the announcement data to an array for later verification. *Ensure thread safety*

vaibhawvipul:
- draft PR


internal/announcements/announce.go
```diff
@@ -0,0 +1,94 @@
...
+			panic("Invalid environment type.")
+		}
+	})
+}
+
+func GetInstance() Announcement {
+	// check if the instance has been initialized
+	if instance == nil {
+		// initialize with the default environment type
+		Initialize(Dst)
```
dtornow: Do not implicitly create an instance but assert that the instance is not null

cmd/dst/run.go
```diff
@@ -70,6 +71,8 @@ func RunDSTCmd() *cobra.Command {
 			}))
 			slog.SetDefault(logger)
 
+			announcements.Initialize(announcements.Dst)
```
dtornow: Add `announcements.Initialize(announcements.Nop)` to production path

internal/app/subsystems/aio/network/network_dst.go
```diff
@@ -69,10 +70,18 @@ func (d *NetworkDSTDevice) Process(sqes []*bus.SQE[t_aio.Submission, t_aio.Compl
 				res = &http.Response{
 					StatusCode: http.StatusOK,
 				}
+
+				event := announcements.NewEvent("HTTPResponse")
```
dtornow, line 74: We need to announce more information about the HttpRequest e.g. url, method, etc to be able to route and assert in model

internal/announcements/announce.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, fmt.Errorf("key '%s' does not exist in event of type '%s'", key, e.Type)
+	}
+	return value, nil
+}
+
+type Announcement interface {
+	Announce(event *Event)
+}
+
+type NopAnnouncement struct{}
```
dfarr: I hate to be that guy but should this be `NoopAnnouncement` ? I know I know, different amount of letters)

dtornow:
A few suggestions:

# Monitors

Make `Monitor` an interface with a method
- `apply(event Event)`
- `status() []error`

Source internal/announcements/monitor.go

# Announcement

Extend the `Announcement` interface with a method
- `register(monitor *Monitor)`

register is a Noop for Noop and adds a monitor to DstAnnouncement.

On announce, iterate over all monitors an apply the event

**Note**

I believe `Announce` and `GetAnnouncements` do not need to be thread safe because they can only be accessed from one thread to begin with

*if this is not true, please discuss, then my mental model is incomplete*

internal/announcements/monitors/taskmonitor.go
```diff
@@ -0,0 +1,37 @@
...
+		eventTypeCount: make(map[string]int),
+	}
+}
+
+func (tm *taskMonitor) Apply(event announcements.Event) {
+	tm.mutex.Lock()
+	defer tm.mutex.Unlock()
+
+	// Increment event type count
+	tm.eventTypeCount[event.Type]++
```
dtornow, line 26: This is the right idea: When an event happens, modify a counter. When we check the monitors at the end of the run, here we want taskCreated == taskCompleted. If that check is too aggressive, at least we want to check that taskCreated >= taskCompleted

```
if event.Type == "TaskCreated" {
  tm.taskCreated ++
}
if event.Type == "TaskCompleted" {
  tm.taskCompleted ++
}
```

internal/announcements/monitor.go
```diff
@@ -0,0 +1,8 @@
+// monitor.go
+
+package announcements
+
+type Monitors interface {
```
dtornow, line 5: I would rename the interface to `Monitor` instead of `Monitors` (to the object not the verb)

internal/app/subsystems/aio/network/network_dst.go
```diff
@@ -69,10 +70,22 @@ func (d *NetworkDSTDevice) Process(sqes []*bus.SQE[t_aio.Submission, t_aio.Compl
...
+				event.Set("StatusCode", res.StatusCode)
+				event.Set("URL", sqe.Submission.Network.Http.Url)
+				event.Set("Method", sqe.Submission.Network.Http.Method)
+				announcements.GetInstance().Announce(event)
 			} else {
 				res = &http.Response{
 					StatusCode: http.StatusInternalServerError,
 				}
+
+				event := announcements.NewEvent("HTTPResponse")
```
dtornow, line 84: Good!!

internal/announcements/monitors/taskmonitor.go
```diff
@@ -0,0 +1,37 @@
+package monitors
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/resonatehq/resonate/internal/announcements"
+)
+
+type taskMonitor struct {
```
dtornow, line 10: TaskMonitor must not part of the core package for announcements and monitors, but part of the test/dst -- it's application specific

dtornow (changes requested):

internal/announcements/announce.go
```diff
@@ -0,0 +1,110 @@
...
+	Announce(event *Event)
+	Register(monitor Monitors)
+}
+
+type NoopAnnouncement struct{}
+
+type DstAnnouncement struct {
+	announcements []Event
+	monitors      []Monitors
+	mutex         sync.Mutex // Mutex for thread safety
```
susarlanikhilesh, line 49: Usually a better practice to have the mutex at the start of the struct. 
As the struct need not calculate the offset every time to find the mutex variable inside the struct.

PR #281, opened 2024-04-10
Test both default and fault injection dst scenarios

dfarr:


dfarr:
Example run: https://github.com/resonatehq/resonate/actions/runs/8638374173

Issue #298, opened 2024-04-24
Add version command


dfarr:
It would be nice to see the version of resonate I have installed locally.
```
resonate --version
```

The version is tied to the git tag and can be set via build flags in our release pipeline, we should assert that the git tag matches the version in our release pipeline.
- [ ] set tag in release pipeline
- [ ] assert git tag matches `resonate --version`

flossypurse:
Version can be derived from the Git Tag, but not necessarily required for the mvp.

rishirishhh:
kindly assign this to me

flossypurse:
@rishirishhh - don't hesitate to reach out for any clarification or to chat about requirements and/or design!

flossypurse:
@rishirishhh - are you still interested in working on this?
If so, I highly recommend joining the Discord: https://resonatehq.io/discord and jump into the resonate-engineering channel to get any help you might need.

fuadnafiz98:
Hi, 
Can I work on this PR?

I already see this opened PR: https://github.com/resonatehq/resonate/pull/342

Is this still going? Can I take inspiration from this code or this solution is not valid?

dfarr:
Just closed that PR for being stale.

I think we can take inspiration from the ldflags, but simplify things a little bit. Here's what I'm thinking.

Add the following to the root command:
```go
var (
	cfgFile string
	version string = "unknown"
)

var rootCmd = &cobra.Command{
	Use:     "resonate",
	Short:   "Resonate: distributed async await",
	Version: version,
}
```

Then we can build resonate with an ldflag like this:
```
go build -ldflags "-X github.com/resonatehq/resonate/cmd.version=SET_VIA_LDFLAGS" -o resonate
./resonate --version
```

Most of the work will involve updating the build scripts to set the version accordingly. What do you think about the following strategy?
1. Use git tag if available (eg: "v1.0.0")
2. Otherwise use commit (eg: "cc66e5135531111757c3e28f0a6fc053bc7c8fa8")
3. Otherwise, fall back to "unknown"

fuadnafiz98:
Hi, 
I think it makes sense. 

I will look into this

PR #301, opened 2024-04-29
add --version option to the rootCmd

N4r35h:
resolves issue #298 

its already built into cobra just needed to add the Version string to the rootCommand

image


guergabo:
Hi @N4r35h, 

Thanks for this! I was thinking of using go's build process for this issue instead and pass the versioning information to the linker via `ldflags`. That way it's not so static and we can configure it externally from a github workflow. Otherwise, we have to modify the source code for each build.

It could default to dev or something of the sorts. 

Feel free to explore this route and update this PR. Thanks.

N4r35h:
sure @guergabo,
i have made the changes please refer to the image and let me know if we should add something else or change things around

i added the vsc.version (automatically set by go build) which is the commit hash also feeling it might be useful where we distribute dev versions of the build for testing and want to correlate to the particular commit also seen it done by the docker cli, we can add more info such as version of golang it was compiled with and built time if required

could you explain about the github release workflow or confirm if thats being used right now ?

`go build -ldflags="-X github.com/resonatehq/resonate/cmd/version.Version=v0.5.0"`

image

cmd/version/version.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"runtime/debug"
+
+	"github.com/spf13/cobra"
+)
+
+var Version string = "dev"
+
+var VersionCmd = &cobra.Command{
+	Use: "version",
```
guergabo: no need to make a whole command out of this a variable should do. check this out: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

cmd/root.go
```diff
@@ -42,6 +43,7 @@ func init() {
 	rootCmd.AddCommand(dst.NewCmd())
 	rootCmd.AddCommand(serve.ServeCmd())
 	rootCmd.AddCommand(quickstart.NewCmd())
+	rootCmd.AddCommand(version.VersionCmd)
```
guergabo: the behavior should still remain using the cobra version flag: `resonate --version`

guergabo (changes requested):

guergabo:
regarding the github release workflow, it starts here: https://github.com/resonatehq/resonate/blob/main/.github/workflows/release.yaml

let me take a look in the morning and suggest some follow up actions, because we would want the version information passed to both our `docker image release build` & `github release artifact build`

guergabo:
ok, for docker image release build:
- most likely `--build-arg` should do the job and use environment variables (the ARG keyword) for this build: https://github.com/resonatehq/resonate/blob/main/Dockerfile#L7

for github release build: 
- we just switched to nix and the ldflags are passed here: https://github.com/resonatehq/resonate/blob/main/flake.nix#L88 
- for this i want to defer to @lucperkins to suggest the best approach to pass an argument.

N4r35h:
thanks for pointing out the flake.nix file

i have made the requested changes consider it a draft let me know if anything else is required
i will look into setting this into the workflow in a while

cmd/root.go
```diff
@@ -17,11 +19,27 @@ import (
 
 var (
 	server, cfgFile string
+	Version         string = "dev"
 )
 
+func computeVersion() string {
```
guergabo, line 25: check out the link i sent before, i'll paste here again: https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications

this whole function is not needed.

guergabo (changes requested):

Issue #169, opened 2023-12-21
Limit schedules queried in schedule promises coroutine


dfarr:
Currently our query looks like
```sql
SELECT
  id, description, cron, promise_id, promise_param_headers, promise_param_data, promise_timeout, last_run_time, next_run_time, created_on, idempotency_key
FROM
  schedules
WHERE
  next_run_time <= ?
```

We should place an ORDER BY and LIMIT on this query. LIMIT will ensure we don't overload our system, and an ORDER BY can ensure that the next batch will pick up where the previous one left off. Additionally, I think we can remove some columns from this query.

We can expose the limit value in the [system configuration](https://github.com/resonatehq/resonate/blob/main/cmd/serve/serve.go#L224-L227).

aneessh18:
this issue can be closed.

PR #307, opened 2024-05-03
add: limit while scheduling promises

aneessh18:
This PR addresses the concerns of this issue https://github.com/resonatehq/resonate/issues/169. 

More information: right now the kernel processes schedules from the databases every 10 ms and creates promises for execution. It finds out the list of promises to be scheduled by executing the READ_ALL_SCHEDULES query against the respective database. This is prone to occur when the system is under high stress as the query can degrade performance when there are a huge number of promises to be scheduled. Instead, this PR ensures that only a limited number of them are scheduled without the system getting overwhelmed. 

Testing
1. I've run the local build and put log statements for recording the number of results in the readSchedules method and I was able to verify that the number of scheduled promises is pegged at the limit. 
2. Unit tests

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -260,12 +260,12 @@ func ServeCmd() *cobra.Command {
 	cmd.Flags().Int("system-notification-cache-size", 100, "max number of notifications to keep in cache")
 	cmd.Flags().Int("system-submission-batch-size", 100, "max number of submissions to process on each tick")
 	cmd.Flags().Int("system-completion-batch-size", 100, "max number of completions to process on each tick")
-	cmd.Flags().Int("system-max-schedules-size", 10000, "max number of schedules to process on each tick")
+	cmd.Flags().Int("system-schedule-batch-size", 10000, "max number of schedules to process on each tick")
 
 	_ = viper.BindPFlag("system.notificationCacheSize", cmd.Flags().Lookup("system-notification-cache-size"))
 	_ = viper.BindPFlag("system.submissionBatchSize", cmd.Flags().Lookup("system-submission-batch-size"))
 	_ = viper.BindPFlag("system.completionBatchSize", cmd.Flags().Lookup("system-completion-batch-size"))
-	_ = viper.BindPFlag("system.maxSchedulesSize", cmd.Flags().Lookup("system-max-schedules-size"))
+	_ = viper.BindPFlag("system.scheduleBatchSize", cmd.Flags().Lookup("system-schedule-batch-size"))
 	// metrics
 	cmd.Flags().Int("metrics-port", 9090, "prometheus metrics server port")
 	_ = viper.BindPFlag("metrics.port", cmd.Flags().Lookup("metrics-port"))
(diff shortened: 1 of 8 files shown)


cmd/serve/serve.go
```diff
@@ -260,11 +260,12 @@ func ServeCmd() *cobra.Command {
 	cmd.Flags().Int("system-notification-cache-size", 100, "max number of notifications to keep in cache")
 	cmd.Flags().Int("system-submission-batch-size", 100, "max number of submissions to process on each tick")
 	cmd.Flags().Int("system-completion-batch-size", 100, "max number of completions to process on each tick")
+	cmd.Flags().Int("system-max-schedules-size", 10000, "max number of schedules to process on each tick")
```
dfarr: Mind renaming this?

```suggestion
	cmd.Flags().Int("system-schedule-batch-size", 10000, "max number of schedules to process on each tick")
```

cmd/serve/serve.go
```diff
@@ -260,11 +260,12 @@ func ServeCmd() *cobra.Command {
...
 
 	_ = viper.BindPFlag("system.notificationCacheSize", cmd.Flags().Lookup("system-notification-cache-size"))
 	_ = viper.BindPFlag("system.submissionBatchSize", cmd.Flags().Lookup("system-submission-batch-size"))
 	_ = viper.BindPFlag("system.completionBatchSize", cmd.Flags().Lookup("system-completion-batch-size"))
-
+	_ = viper.BindPFlag("system.maxSchedulesSize", cmd.Flags().Lookup("system-max-schedules-size"))
```
dfarr:
```suggestion
	_ = viper.BindPFlag("system.scheduleBatchSize", cmd.Flags().Lookup("system-max-schedules-size"))
```

internal/kernel/system/system.go
```diff
@@ -19,6 +19,7 @@ type Config struct {
 	NotificationCacheSize int
 	SubmissionBatchSize   int
 	CompletionBatchSize   int
+	MaxSchedulesSize      int
```
dfarr:
```suggestion
	ScheduleBatchSize      int
```

dfarr (changes requested):
Thank you so much again @aneessh18!!!

Do you mind adding the schedule batch size configuration to our deterministic simulation testing (DST) flags? We would need to add it [here](https://github.com/resonatehq/resonate/blob/main/cmd/dst/run.go#L269-L275), the range flag generates a random number for the config value between the min and the max. It would also be useful to print out this config value [here](https://github.com/resonatehq/resonate/blob/main/internal/kernel/system/system.go#L24-L31), this helps compare two runs of our DST to ensure that the configuration value is the same.

You can try the DST locally!
```
go run ./... dst run
```

aneessh18:
> Thank you so much again @aneessh18!!!
> 
> Do you mind adding the schedule batch size configuration to our deterministic simulation testing (DST) flags? We would need to add it [here](https://github.com/resonatehq/resonate/blob/main/cmd/dst/run.go#L269-L275), the range flag generates a random number for the config value between the min and the max. It would also be useful to print out this config value [here](https://github.com/resonatehq/resonate/blob/main/internal/kernel/system/system.go#L24-L31), this helps compare two runs of our DST to ensure that the configuration value is the same.
> 
> You can try the DST locally!
> 
> ```
> go run ./... dst run
> ```

Thanks for reviewing this MR. Added the flags in DST.

cmd/serve/serve.go
aneessh18: Done

dfarr (approved):
🎉

PR #308, opened 2024-05-03
feat(timers): enabling simple timers

guergabo:
### Changes 

This PR introduces a new `resonate:timeout` tag. 

### Context

The new tag enables distinct behaviors between vanilla promises and simple timeout promises: 

- When a vanilla promise timeouts, its state will be updated to REJECTED_TIMEDOUT
- When a simple timeout promise timeouts, its state will be updated to RESOLVED

This distinction will allow events to be tied to the timeout of a promise.

### DIFF
--- a/pkg/promise/promise.go
+++ b/pkg/promise/promise.go
@@ -107,3 +107,14 @@ func (v Value) String() string {
 		string(v.Data),
 	)
 }
+
+func GetTimedoutState(p *Promise) State {
+	completedState := Timedout
+	if v, ok := p.Tags["resonate:timeout"]; ok {
+		if v == "true" {
+			completedState = Resolved
+		}
+	}
+
+	return completedState
+}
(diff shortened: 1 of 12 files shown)


pkg/promise/promise.go
```diff
@@ -39,11 +39,11 @@ func (p *Promise) String() string {
 type State int
 
 const (
-	Pending State = 1 << iota
-	Resolved
-	Rejected
-	Canceled
-	Timedout
+	Pending  State = 1 << iota // 1
```
dfarr, line 42: Thank you!!!

dfarr:
Okay hear me out, not sure if this is possible...

What if we define a stored column (similar to `invocation`) and cast it to an integer corresponding to the `State` enum. If there is no tag set, we default this value to 16 (timedout). This way, we can always just refer to this column to set the state of the promise on timeout. On an API level we could validate that this value is either {resovled,timedout} (and map those to their corresponding integers), but this would also future proof our server with the ability to transition to _any_ state on timedout if desired. What do you think?

guergabo:
Going the tag route, move invocation to tags.

guergabo:
Ready for review.

dfarr (approved):

Issue #112, opened 2023-11-14
Always panic through utils.Assert


dfarr:
Sometimes when we encounter an intolerable error we call `panic()` directly, instead we could call `utils.Assert(false, ...)` that way we would have the opportunity to optionally turn off strict panicing in production.

guergabo:
this would be nice to expose as a flag like `--strict`

guergabo:
@KannappanChida. of course. lmk if you need any help. thank you !

guergabo:
interesting approach at sqlite. They have different behaviors for asserts depending on the build: https://www.sqlite.org/assert.html#:~:text=In%20SQLite%2C%20the%20presence%20of,about%20the%20truth%20of%20X.

flossypurse:
@KannappanChida - curious if you are still interested in working on this?

PR #329, opened 2024-05-16
add strict flag for assert

susarlanikhilesh:
Addresses [Issue](https://github.com/resonatehq/resonate/issues/112)


dfarr:
Thank you so much @susarlanikhilesh 🎉 🎉 

The tests are failing because some of our tests expect a panic to occur, and if the flag is not set (eg: the code is run via a test rather than cli) I believe viper is returning the default boolean value of `false`. Is there a way to check if the flag is not set and, if not, set strict to true instead?

susarlanikhilesh:
Yes the strict flag is false, I did set when panic is true in the test cases

dfarr:
I would prefer for `strict=true` in cases where not explicitly set. Our philosophy here is to panic by default when an unexpected event occurs, this way we can protect ourselves from being in an invalid state.

```
viper.GetBool("strict") // if not set needs to be defaulted to true
```

An easier way to do it might be to invert the name of the flag to something like `ignore-asserts` that we can default to false which makes things a little easier to implement.

internal/util/util.go
```diff
@@ -6,10 +6,12 @@ import (
 	"time"
 
 	"github.com/robfig/cron/v3"
+	"github.com/spf13/viper"
 )
 
 func Assert(cond bool, msg string) {
-	if !cond {
+	strict := viper.GetBool("ignore-asserts")
```
dfarr:
```suggestion
	ignoreAsserts := viper.GetBool("ignore-asserts")
```

internal/util/util.go
```diff
@@ -6,10 +6,12 @@ import (
...
+	if !strict && !cond {
```
dfarr:
```suggestion
	if !ignoreAsserts && !cond {
```

dfarr (changes requested):

dfarr (approved):

Issue #330, opened 2024-05-17
Basic Auth


dtornow:
## Describe the problem you are facing 

Currently, the Resonate Server does not support authentication and authorization

## Describe the solution you'd like

Add Basic Auth for authentication for the Resonate's http api

## Additional context

This is a first step towards a mature authentication and authorization system. The design of the authentication and authorization system will be discussed in a separate issue

dtornow:
Resonate uses the [gin web framework](https://gin-gonic.com) so we can use the gin Basic Auth middleware. The (list of) user can be configured via the configuration file or command line arguments. E.g.

```
// Apply Basic Auth middleware conditionally
    authorized := r.Group("/")
    // username password pairs could also be a list to configure multiple accounts
    if config.AuthUsername != "" && config.AuthPassword != "" {
        accounts := gin.Accounts{
            config.AuthUsername: config.AuthPassword,
        }
        basicAuthMiddleware := gin.BasicAuth(accounts)
        authorized.Use(basicAuthMiddleware)
    }

    // Promises API
    authorized.POST("/promises", s.createPromise)
    authorized.GET("/promises", s.searchPromises)
    authorized.GET("/promises/*id", s.readPromise)
    authorized.PATCH("/promises/*id", s.completePromise)
    
    // ...
```

See [source file](https://github.com/resonatehq/resonate/blob/9f31bbeea4951acf642553fae4a533b7c1ab2624/internal/app/subsystems/api/http/http.go#L28)

In subsequent iterations, we need to design how to attach authenticated user information to our api requests.

dtornow:
**Additional information**

- Every API endpoint is registered on the authorized group

- `Config` needs to be extended to contain `AuthUsername` and `AuthPassword` (if we only care about single user) or a list of Username Password pairs (if we want to support multiple users)

- Username and password can be configured via config file or via cobra command line arguments

RyanCarlisle:
Hi, is this issue already assigned to someone?

hkiiita:
@dtornow requesting your review on https://github.com/resonatehq/resonate/pull/333

PR #333, opened 2024-05-20
Addition of basic Auth functionality as mentioned in issue Basic Auth #330 

hkiiita:
Addition of basic Auth functionality as mentioned in issue Basic Auth #330

### DIFF
--- a/internal/creds/processCreds.go
+++ b/internal/creds/processCreds.go
@@ -1,0 +1,34 @@
+package creds
+
+import (
+	"errors"
+	"log/slog"
+)
+
+type Credential struct {
+	Username string `yaml:"username"`
+	Password string `yaml:"password"`
+}
+
+type CredentialsList struct {
+	Users []Credential `yaml:"users"`
+}
+
+var CredsFromFile CredentialsList
+
+func GetCredentials() (CredentialsList, error) {
+	if len(CredsFromFile.Users) == 0 {
+		slog.Error("CredentialsList is empty", "error", errors.New("Credentials are empty."))
+		return CredentialsList{}, errors.New("Credentials are empty.")
+	} else {
+		return CredsFromFile, nil
+	}
+}
+
+func GetProcessedCreds(credList CredentialsList) map[string]string {
+	credentials := make(map[string]string)
+	for _, item := range credList.Users {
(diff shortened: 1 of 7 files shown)


hkiiita:
Incorporates multiple users.

internal/creds/processCreds.go
```diff
@@ -0,0 +1,34 @@
...
+	Users []Credential `yaml:"users"`
+}
+
+var CredsFromFile CredentialsList
+
+func GetCredentials() (CredentialsList, error) {
+	if len(CredsFromFile.Users) == 0 {
+		slog.Error("CredentialsList is empty", "error", errors.New("Credentials are empty."))
+		return CredentialsList{}, errors.New("Credentials are empty.")
+	} else {
```
susarlanikhilesh, line 23: else is not required

dfarr:
Thank you so much for the PR @hkiiita 🎉 

We can't believe it, but we actually got two PRs for the same feature. We are going to merge #332, but would you be up for incorporating multiple users (which is lacking in #332) once that one is merged?

hkiiita:
> Thank you so much for the PR @hkiiita 🎉
> 
> We can't believe it, but we actually got two PRs for the same feature. We are going to merge #332, but would you be up for incorporating multiple users (which is lacking in #332) once that one is merged?

Hi @dfarr thank you so much for your kind reply on this. Yes , i will be more than happy to continue to work on this after 332 gets merged. I have already incorporated multiple users authentication into my PR #333 , so will wait on your side for a review and further steps. Thanks again !

dfarr:
Hey @hkiiita, PR #332 has been merged, if you're still up for it, we could extend basic auth to support multiple users - I believe we would want to extend [these arguments](https://github.com/resonatehq/resonate/blob/main/cmd/serve/serve.go#L193-L194) to take an array of username/password combos (and propagate that array down accordingly)

hkiiita:
Hi @dfarr , thanks for the reply . Yes I definitely would want to work on this . Just wanted to share that, if you are ok with it ,  I will reach out to you on 18th June.

hkiiita:
https://github.com/resonatehq/resonate/pull/352

PR #340, opened 2024-05-28
feat(db-migrations): add basic support for db migrations

guergabo:
x

### DIFF
--- a/internal/app/subsystems/aio/store/migrations.go
+++ b/internal/app/subsystems/aio/store/migrations.go
@@ -43,6 +43,8 @@ func Start(db *sql.DB, txTimeout time.Duration, migrationsFS embed.FS) error {
 				err = fmt.Errorf("tx failed: %v, unable to rollback: %v", err, rbErr)
 			}
 		}
+
+		err = fmt.Errorf("tx failed, performed a rollback: %v", err)
 	}()
 
 	for _, m := range migrationsToApply {
--- a/internal/app/subsystems/aio/store/migrations.go
+++ b/internal/app/subsystems/aio/store/migrations.go
@@ -13,22 +13,146 @@ import (
 	"time"
 )
 
-type migration struct {
+type Plan int
+
+const (
+	Default Plan = iota // 0 is the default value
+	DryRun
+	Apply
+)
+
+type Migration struct {
 	Version int
 	Content []byte
 }
 
-func Start(db *sql.DB, txTimeout time.Duration, migrationsFS embed.FS) error {
-	currentVersion, err := readVersion(db)
+type MigrationPlan []Migration
+
+func (m MigrationPlan) String() string {
+	var sb strings.Builder
+	sb.WriteString("Operations to perform:\n")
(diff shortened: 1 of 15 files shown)


internal/app/subsystems/aio/store/postgres/migrations/schema-0000.sql
```diff
@@ -0,0 +1,128 @@
...
+    promise_id   TEXT,
+    url          TEXT,
+    retry_policy BYTEA,
+    time         BIGINT,
+    attempt      INTEGER,
+    PRIMARY KEY(id, promise_id)
+);
+
+-- CreateTable
+CREATE TABLE IF NOT EXISTS migrations (
```
dfarr: I wonder if this table should be created outside of the schema file, that way the migrations module can "own" the table vs the sql/postgres module (which owns the rest of the tables). One advantage of this approach would be knowing the table must exist before querying for the latest version and our code could assume that no rows returned == fresh db.

internal/app/subsystems/aio/store/postgres/postgres.go
```diff
@@ -470,11 +370,7 @@ func (s *PostgresStore) String() string {
 }
 
 func (s *PostgresStore) Start() error {
-	if _, err := s.db.Exec(CREATE_TABLE_STATEMENT); err != nil {
-		return err
-	}
-
-	return nil
+	return store.Start(Version, s.db, 10*time.Second, migrationsFS, s.config.Plan)
```
dfarr: Could we rename this function to something to indicate this is a migration? Another possibility would be to add a new function called `Migrate` and have start call that function, then the migrate command could also call that function directly and avoid the start/stop.

internal/app/subsystems/aio/store/migrations.go
```diff
@@ -0,0 +1,186 @@
...
+
+	switch plan {
+	case Default:
+		return fmt.Errorf("database version %d does not match current version %d please run `resonate migrate --plan` to see migrations needed", dbVersion, currVersion)
+	case DryRun:
+		plan, err := generateMigrationPlan(migrationsFS, dbVersion)
+		if err != nil {
+			return err
+		}
+		fmt.Println("Migrations to apply:")
```
dfarr: Can we use slog for this print statements?

internal/app/subsystems/aio/store/migrations.go
```diff
@@ -0,0 +1,186 @@
...
+	}()
+
+	for _, m := range plan {
+		_, err = tx.Exec(string(m.Content))
+		if err != nil {
+			return fmt.Errorf("failed to execute migration version %d: %w", m.Version, err)
+		}
+	}
+
+	if err = tx.Commit(); err != nil {
```
dfarr: Should we also write the version to the migrations table as part of this transaction explicitly here?

guergabo:
Ready for review. Before merging have to remove both migrations file, but keeping for review as example of what they would look like.

PR #352, opened 2024-06-19
Basic Auth for multiple users.

hkiiita:
This PR adresses Basic Auth #330 for multiple users. Please review.


hkiiita:
@dfarr Please have a look.

dfarr:
Sorry for the delay in reviewing @hkiiita , this looks great! I have one request, rather than have a separate file for username/password information could we change this to use the default configuration file? The file could look like this:
```yaml
auth:
  - username: user1
    password: pass1
  - username: user2
    password: pass2
```

hkiiita:
Thanks for having a look @dfarr ; please review the new changes. I have removed the separate creds file as asked and am now reading credentials from the main config file itself.

cmd/root.go
```diff
@@ -67,6 +70,13 @@ func initConfig() {
 			os.Exit(1)
 		}
 	}
+
+	// Unmarshal the 'auth' key into the slice of Auth structs
+	err := viper.UnmarshalKey("auth", &creds.CredsFromFile)
```
dfarr: I think we should be able to remove this now

cmd/serve/serve.go
```diff
@@ -56,6 +57,13 @@ func ServeCmd() *cobra.Command {
 			api := api.New(config.API.Size, metrics)
 			aio := aio.New(config.AIO.Size, metrics)
 
+			credentials, err := creds.GetCredentials()
```
dfarr: I believe viper automatically parses configuration from the file (and command line arguments) we shouldn't need to manually parse/set `config.API.Subsystems.Http.Auth` like this

dfarr (changes requested):
Thank you @hkiiita! Now that we are using viper for the configuration we shouldn't need to do manual parsing of the configuration, as long as we specify the type I believe viper will parse automatically for us.

internal/creds/processCreds.go
```diff
@@ -0,0 +1,34 @@
...
+type CredentialsList struct {
+	Users []Auth `yaml:"auth"`
+}
+
+var CredsFromFile CredentialsList
+
+func GetCredentials() (CredentialsList, error) {
+	if len(CredsFromFile.Users) == 0 {
+		slog.Error("CredentialsList is empty", "error", errors.New("Credentials are empty."))
+		return CredentialsList{}, errors.New("Credentials are empty.")
```
dfarr: If the list is empty this can be the signal to disable basic auth. It looks like returning an error here is causing the [linearizability check](https://github.com/resonatehq/resonate/actions/runs/9741229922/job/26907803484#step:6:8) to fail

cmd/root.go
hkiiita: Thanks for reviewing , yup removed it.

cmd/serve/serve.go
hkiiita: Yes, changed it accordingly.

internal/creds/processCreds.go
hkiiita: Yes.

cmd/serve/serve.go
```diff
@@ -56,6 +56,8 @@ func ServeCmd() *cobra.Command {
 			api := api.New(config.API.Size, metrics)
 			aio := aio.New(config.AIO.Size, metrics)
 
+			config.API.Subsystems.Http.Auth = viper.Get("auth").([]interface{})
```
dfarr: We can remove this

internal/app/subsystems/api/http/http.go
```diff
@@ -2,28 +2,22 @@ package http
...
 
-type Auth struct {
-	Username string
-	Password string
-}
-
 type Config struct {
 	Addr    string
-	Auth    *Auth
+	Auth    []interface{}
```
dfarr:
```suggestion
	Auth    map[string]string
```

I was doing some experimenting and I think using the the map type that gin expects will make our life a lot easier :)

internal/app/subsystems/api/http/http.go
```diff
@@ -48,16 +50,7 @@ func New(api api.API, config *Config) api.Subsystem {
...
-		util.Assert(config.Auth.Username != "", "http basic auth username is required")
-		util.Assert(config.Auth.Password != "", "http basic auth password is required")
-
-		accounts := gin.Accounts{
-			config.Auth.Username: config.Auth.Password,
-		}
-		basicAuthMiddleware := gin.BasicAuth(accounts)
-		authorized.Use(basicAuthMiddleware)
-	}
+	authorized.Use(BasicAuthMiddleware(config))
```
dfarr: I've noticed that specifying a nil value for `Use` leads to runtime issues so I think we need to conditionally set the middleware
```
if len(config.Auth) > 0 {
    authorized.Use(gin.BasicAuth(config.Auth))
}
```

internal/app/subsystems/api/http/http.go
dfarr: In order to do this we need to make two small tweaks to the flags defined in `cmd/serve/serve.go`. We need to replace these flags:
- api-http-auth-username / api.subsystems.http.auth.username
- api-http-auth-password / api.subsystems.http.auth.password

With a single map flag
```go
cmd.Flags().StringToString("api-http-auth", map[string]string{}, "basic auth username/password pairs")
_ = viper.BindPFlag("api.subsystems.http.auth", cmd.Flags().Lookup("api-http-auth"))
```

dfarr (changes requested):

internal/app/subsystems/api/http/http.go
hkiiita: Yup, changed the type as indeed we would need to pass a map to gin , and added cobra flags as suggested by you , as i undertsand we do also have to keep the option to provide credentials via the command line too. Also we need to bind  with viper because viper needs to be the single source of truth for 'getting' almost all values, both from a config file as well as those that get passed on command line :-)

internal/app/subsystems/api/http/http.go
hkiiita: Indeed. I also removed the function ```BasicAuthMiddleware``` as i thought it wasnt needed at this juncture after new changes. Also, i learned that the length check here would not be an issue as we would already have our ```config``` initialised. Thus , in a scenario where no creds are provided , the ```config.Auth```  would still hold an initialised empty map .

cmd/serve/serve.go
hkiiita: Yup, removed it . Thanks for being with me @dfarr .

Now , i have a doubt , as am still learning, I saw that in ``` cmd/serve/serve.go``` we call ```config, err := util.NewConfig()``` which basically calls ```viper.Unmarshal(&config)``` on the config file and will fill up our Config struct and all its fields and sub-fields. Essentially picking up the creds present there.

Now, in a scnenario where we do have creds present in the config file but the user also provides additional creds on commandline as well using the ```api-http-auth``` flag , or , if there are no creds in config file but are only provided via the command line; in that case how would our ```Config.Auth``` get filled ? Would we need a function to be created which looks for a ```api-http-auth``` flag being invoked and accordigly initialises ```config.Auth``` with the values recieved from flag ?

dfarr: Good question! Command line arguments take precedent over the config file, which is what we want. You can test all this out locally by running serve with differing command line arguments/config file values.
```
go run ./... serve
```

To interact with the server you can use the cli
```
go run ./... promises get foo
go run ./... promises get foo --username foo --password foo
```

internal/app/subsystems/api/http/http_test.go
```diff
@@ -56,24 +56,20 @@ func (t *httpTest) teardown() error {
...
-		{
-			name: "NoAuth",
-			auth: &Auth{},
-		},
 		{
 			name:        "BasicAuthCorrectCredentials",
-			auth:        &Auth{Username: "username", Password: "password"},
-			reqUsername: "username",
-			reqPassword: "password",
+			auth:        map[string]string{"username": "user1", "password": "pass1"},
```
dfarr: Whoops, this map is supposed to be
```
{"username": "password"}
```

And the reqUsername and reqPassword in this one needs to match the configured password. I think this should resolve the github action failure

cmd/serve/serve.go
hkiiita: I ran it and loved to see how things were working seamlessly. Passed dummy creds to server , then Created a dummy promise ID and made these ```get``` requests by providing credentials on command line. Loved seeing the appropriate responses.

Learned the proper flow and usage of cobra commands and the logic of binding flags to viper .


Thanks for the guidance @dfarr

dfarr (approved):
🎉

hkiiita:
> 🎉

Thank you so much @dfarr for the journey. I definitely learned a couple of things during this. Thanks again !

Issue #387, opened 2024-08-21
bad CPU type in executable on Apple M3 Pro arm64 arch


flossypurse:
## Expected Behavior

Running the following commands should start the Resonate Server on MacOS:

```
brew install resonatehq/tap/resonate
resonate serve
```

## Actual Behavior

The `resonate serve` command causes this error:

```
zsh: bad CPU type in executable: resonate
```

## To Reproduce
  1. Use an Apple M3 Pro laptop with arm64 architecture
  2. Brew install resonate
  3. Run the `resonate serve` command

flossypurse:
Possible need to change something in the homebrew repo?
https://github.com/resonatehq/homebrew-tap

PR #417, opened 2024-09-13
Update release workflow to support arm64

dfarr:
The release was not being built on arm64 chip architecture for both linux and macos. Github now supports arm64 for macos by default, additionally we upgraded our github account to get access to a linux arm64 runner.

I added a step to verify the executable targets the correct architecture and expanded the verify step to run against all runners/artifacts.

### DIFF
--- a/.github/workflows/release_publish_github_artifacts.yaml
+++ b/.github/workflows/release_publish_github_artifacts.yaml
@@ -21,26 +21,30 @@ jobs:
         systems:
           - os: linux
             arch: x86_64
+            arch_grep: x86-64
             runner: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


dfarr:
- [successful run in a forked repo](https://github.com/resonatehq/resonate2/actions/runs/10839897564)
- [release with artifacts](https://github.com/resonatehq/resonate2/releases/tag/v0.6.0) (can be downloaded and verified)
- [new github runner](https://github.com/organizations/resonatehq/settings/actions/github-hosted-runners/5?viewing_from_runner_group=false)

dfarr:
Resolves #387

PR #437, opened 2024-10-04
Remove required for timeout and ttl fields

dfarr:
The struct tag binding "required" excludes setting 0, which is valid for both ttl and timeout.


Tomperez98 (approved):
Nice!

PR #440, opened 2024-10-04
Ignore 429 and make sure msgs have been processed

### LINKED ISSUE #439: Non determinism in poll plugin tests
The new poll plugin unit tests are failing non-deterministically. This is likely due to not ensuring the http test server is up before making requests. So far, re-running the tests generally will succeed but we should fix the underlying issue.

I suspect this issue is [here](https://github.com/resonatehq/resonate/blob/main/internal/app/plugins/poll/poll_test.go#L182), it is possible to make this request _before_ the server is accepting incoming connections.

Tomperez98:
```go
// internal/app/plugins/poll
if !h.Connect(conn) {
		http.Error(w, "too many requests", http.StatusTooManyRequests)
		return
	}
```

The issue is caused when the poll worker is not up fast enough to process the first three msgs on the test before the forth one gets sent. When this happen we were not checking for `error 429` and doing something on the test side.

The PR continues the iteration when the error is `429`. and ensures all msgs are processed before continuing with the test

### DIFF
--- a/internal/app/plugins/poll/poll_test.go
+++ b/internal/app/plugins/poll/poll_test.go
@@ -213,12 +213,16 @@ func TestPollPlugin(t *testing.T) {
 
 			// establish connections
 			for _, conn := range tc.connections {
-				disconnected.Add(1)
-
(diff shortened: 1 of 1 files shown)


dfarr (approved):
🎉

Issue #436, opened 2024-10-03
Add task cli commands


dfarr:
Claim, complete, and heartbeat tasks.

Commands:
- claim
- complete
- heartbeat

Examples:
```
resonate tasks claim --id foo --counter 1 --process-id bar --ttl 1m
resonate tasks complete --id foo --counter 1
resonate tasks heartbeat --id foo --counter 1
```

See [promises](https://github.com/resonatehq/resonate/tree/main/cmd/promises) and [schedules](https://github.com/resonatehq/resonate/tree/main/cmd/schedules) for a reference.

TusharMohapatra07:
Hello, does this issue needs any prerequisite knowledge of cloud ?

dfarr:
> Hello, does this issue needs any prerequisite knowledge of cloud ?

Not at all! This is just request/response from client to server.

TusharMohapatra07:
Okay then, i'd like to work on this.

dfarr:
Thank you @TusharMohapatra07 !!!

dfarr:
Completed by #443

PR #443, opened 2024-10-17
add claim, complete and heartbeat tasks to cli

TusharMohapatra07:
- This PR is linked to issue #436 . I've implemented the claim and complete task.
- Need some help in the following areas:
  1. found an endpoint for hearbeatTasks.go (https://github.com/resonatehq/resonate/blob/main/internal/app/coroutines/heartbeatTasks.go) but according to the documentation (https://resonatehq.github.io/resonate/#tag/Tasks/operation/completeTask), its a post request requiring just the process-id. Unable to find the endpoint to map id and counter with
  2. If this implementation is correct, I need to pass a type that implements `gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any]`. In tasks.go i need to register these commands in order for them to work. I've so far figured out that gocoro.newCoroutine with signature `func[T, TNext, TReturn any](f CoroutineFunc[T, TNext, TReturn], r map[string]any) *coroutine[T, TNext, TReturn]` creates a coroutine that implements this type. But i'm unable to figure out what to pass in place of f and r

### DIFF
--- a/cmd/tasks/claim.go
+++ b/cmd/tasks/claim.go
@@ -1,0 +1,72 @@
+package tasks
+
+import (
+	"time"
+
+	"github.com/resonatehq/gocoro"
+	"github.com/resonatehq/resonate/internal/app/coroutines"
+	"github.com/resonatehq/resonate/internal/kernel/t_aio"
+	"github.com/resonatehq/resonate/internal/kernel/t_api"
+	"github.com/spf13/cobra"
+)
+
+var claimTaskExample = `
+# Claim a task
+resonate tasks claim --id foo --counter 1 --process-id bar --ttl 1m`
+
+// ClaimTask sets up the CLI command for claiming a task using provided task information.
+// It takes a gocoro.Coroutine as an argument which allows asynchronous task handling.
+func ClaimTask(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any]) *cobra.Command {
+
+	var (
+		// Command-line flags for user input
+		id        string
+		counter   int
+		processID string
+		ttl       time.Duration
+	)
+
+	// Define the cobra command for `claim`
+	cmd := &cobra.Command{
+		Use:     "claim <id> <counter> <processID> <ttl>",
+		Short:   "Claim Task",
+		Example: claimTaskExample,
+		Run: func(cmd *cobra.Command, args []string) {
+			// Build the request for claiming a task with the user's input
+			req := &t_api.Request{
+				ClaimTask: &t_api.ClaimTaskRequest{
+					Id:        id,
+					Counter:   counter,
+					ProcessId: processID,
+					Ttl:       int(ttl.Seconds()),
+				},
+			}
+
+			// Build the request for claiming a task
+			resp, err := coroutines.ClaimTask(c, req)
+
+			// Handle errors during task claiming
+			if err != nil {
+				cmd.PrintErrf("Error claiming task: %v\n", err)
+				return
+			}
+
+			// If successful, display the claimed task information
+			cmd.Printf("Claimed Task: %+v\n", resp.ClaimTask)
+		},
+	}
+
+	// Define flags that allow users to pass values from the command line
+	cmd.Flags().StringVar(&id, "id", "", "Task ID")
+	cmd.Flags().IntVar(&counter, "counter", 1, "Task counter")
+	cmd.Flags().StringVar(&processID, "process-id", "", "Process ID")
+	cmd.Flags().DurationVar(&ttl, "ttl", 1*time.Minute, "Time to live (e.g: 1m, 30s)")
+
+	// Mark certain flags as required so the user must provide them
+	cmd.MarkFlagRequired("id")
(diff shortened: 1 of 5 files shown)


dfarr:
Thank you @TusharMohapatra07!!!

The scaffolding here looks great, exactly what we need. But we do need to use the openapi auto generated client here, not the `Request`/`Response` objects which are internal facing structures for the server only. Here the cli is acting as a client and getting a response back from the server.

The promise commands and schedule commands should be good guides, for example let's take a look at [promise create](https://github.com/resonatehq/resonate/blob/main/cmd/promises/create.go).

1. Pass the client in to the command as a [function parameter](https://github.com/resonatehq/resonate/blob/main/cmd/promises/create.go#L21)
2. Take all the command line arguments and construct a params/body struct to pass to the [generated function](https://github.com/resonatehq/resonate/blob/main/cmd/promises/create.go#L69) (there will be functions corresponding to tasks too).
3. Grab the response and print a corresponding status message

You may find our api specifiction docs [here](https://resonatehq.github.io/resonate/#tag/Tasks/operation/claimTask) handy too. Let me know if I can help further!

TusharMohapatra07:
@dfarr So, i need to use the functions present in github.com/resonatehq/resonate/pkg/client/v1. correct ?

dfarr:
> @dfarr So, i need to use the functions present in github.com/resonatehq/resonate/pkg/client/v1. correct ?

Correct!

TusharMohapatra07:
@dfarr on it.

TusharMohapatra07:
@dfarr please review and suggest changes (if any).

cmd/tasks/claim.go
```diff
@@ -0,0 +1,94 @@
...
+			params := &v1.ClaimTaskParams{
+				RequestId: &requestId, // Optional tracking ID
+			}
+
+			// Create the body for the claim task request
+			body := v1.ClaimTaskJSONRequestBody{
+				Id:        id,
+				Counter:   counter,
+				ProcessId: processId,
+				Ttl:       int64(ttl.Milliseconds()), // Convert duration to milliseconds
```
dfarr: I believe `ttl.Milliseconds()` is already an int64 so we shouldn't have to convert it

cmd/tasks/claim.go
```diff
@@ -0,0 +1,94 @@
...
+			res, err := c.V1().ClaimTaskWithResponse(context.TODO(), params, body)
+
+			if err != nil {
+				return err // Return any errors from the request
+			}
+
+			// Handle the response based on the status code
+			if res.StatusCode() == 201 {
+				cmd.Printf("Task claimed: %s\n", id)
+			} else if res.StatusCode() == 403 {
```
dfarr: For consistency, I think we should [print the entire body](https://github.com/resonatehq/resonate/blob/main/cmd/promises/get.go#L37-L40) if the status code is not 201, this is what we do for promises and schedules and will keep our cli more flexible to changes. Could we follow the same pattern for all task commands?

dfarr (changes requested):
This looks really good! I have a few small requests.

One additional thought: for `promises` and `schedules` the id parameter is a positional parameter (see [here](https://github.com/resonatehq/resonate/blob/main/cmd/promises/get.go#L22-L30)), I'm wondering if we should take a similar approach for tasks to be consistent, what do you think?

TusharMohapatra07:
@dfarr I think having id as positional parameter consistently throughout the application enhances the user experience a lot. i'm implementing those changes you requested.

TusharMohapatra07:
@dfarr please review. Also, the success code for claim and complete is 201 and that of heartbeat is 200 (according to docs). So, i've printed the response body when the status code is not equal to the above mentioned.

dfarr:
Very good! Only one thing, the heartbeat request takes a single parameter of `--process-id`, whichever process id is specified on claim can be used to heartbeat, this way a client can claim a bunch of tasks and batch heartbeat with a single process id.

```
resonate tasks heartbeat --process-id foo
```

It must be hard to test this! Here is how I do it, maybe this will help
```
# start the server
go run ./... serve

# create a promise and a task
go run ./... promises create foo --timeout 24h --tag resonate:invoke=poll://default

# claim the task (the id is auto incrementing, starts with 1)
go run ./... tasks claim 1 --counter 1 --process-id foo --ttl 1m

# complete the task (if not completed in one minute, will need to be re-claimed, the counter will be bumped)
go run ./... tasks complete 1 --counter 1
```

If you have a sqlite db editor you can open up the db file (defaults to `resonate.db`) and inspect the tasks table to see how the cli command affects tasks.

One more nitpick, we don't need the request id for the cli. It doesn't hurt to keep it so I am of two minds on this one, but getting rid of it as a flag might make the cli a little simpler.

TusharMohapatra07:
@dfarr I'll get rid of requestId. Sorry but i'm not clear about the first requirement. What should i do with the process-id ?   I'm not clear what changes i need to implement.

TusharMohapatra07:
@dfarr please review

dfarr (commented):
🎉

PR #446, opened 2024-10-24
Added Callback cli

### LINKED ISSUE #435: Add callback cli command
Create a callback associated with a promise.

Commands:
- create

Examples:
```
resonate callbacks create --promise-id foo --root-promise-id bar --timeout 1h --recv default
```

A receiver can be either a string or an object, an object must always have a type and data.
- logical as string (eg: "default")
- physical as url string (eg: "poll://default/6fa89b7e-4a56-40e8-ba4e-78864caa3278")
- physical as object (eg: `{"type": "poll", "data": {"group": "default", "id": "6fa89b7e-4a56-40e8-ba4e-78864caa3278"}}`

See [promises](https://github.com/resonatehq/resonate/tree/main/cmd/promises) and [schedules](https://github.com/resonatehq/resonate/tree/main/cmd/schedules) for a reference.

Madraceee:
### Description
Added callback cli. Invokes the inbuilt commands.

**This is not finished**

1. What should the value of requestID be. UUID or some string? 
2. For url and default, how should those values be passed?

Sry I was busy with my work, could not look into what should be passed for callbacks which dont require a json

**To-Do**
- [ ] Fix RequestID
- [ ] Fix non json flow

### DIFF
--- a/cmd/callbacks/create.go
+++ b/cmd/callbacks/create.go
@@ -1,0 +1,103 @@
+package callbacks
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+	"time"
+
+	"github.com/resonatehq/resonate/pkg/client"
+	v1 "github.com/resonatehq/resonate/pkg/client/v1"
+	"github.com/spf13/cobra"
+)
+
+var createCallbacksExample = `
+# Create a callback
+resonate callback create --promise-id foo --root-promise-id bar --timeout 1h --recv default
+
+# Create a callback with url
+resonate callback create --promise-id foo --root-promise-id bar --timeout 1h --recv poll://default/6fa89b7e-4a56-40e8-ba4e-78864caa3278
+
+# Create a callback with object
+resonate callback create --promise-id foo --root-promise-id bar --timeout 1h --recv {"type": "poll", "data": {"group": "default", "id": "6fa89b7e-4a56-40e8-ba4e-78864caa3278"}}
+`
+
+func CreateCallbackCmd(c client.Client) *cobra.Command {
+	var (
+		promiseId     string
+		rootPromiseId string
+		timeout       time.Duration
+		recv          string
+	)
+	cmd := &cobra.Command{
+		Use:     "create ",
+		Short:   "Create callbacks",
+		Example: createCallbacksExample,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			paramsByte, err := json.Marshal(struct {
+				RequestId string `json:"request-id"`
+			}{RequestId: ""})
+
+			if err != nil {
+				return nil
+			}
+
+			paramsStr := string(paramsByte)
+			params := &v1.CreateCallbackParams{
+				RequestId: &paramsStr,
+			}
+
+			decoder := json.NewDecoder(strings.NewReader(recv))
+			var recvJson v1.Recv
+
+			if err := decoder.Decode(&recvJson); err == nil {
+				body := v1.CreateCallbackJSONRequestBody{
+					PromiseId:     promiseId,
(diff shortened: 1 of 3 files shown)


Madraceee:
@dfarr I couldnt find an option to request for review , so can you help me out?

cmd/callbacks/create.go
```diff
@@ -0,0 +1,104 @@
...
+		recv          string
+	)
+	cmd := &cobra.Command{
+		Use:     "create ",
+		Short:   "Create callbacks",
+		Example: createCallbacksExample,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// What should requestID be?
+			paramsByte, err := json.Marshal(struct {
```
dfarr, line 39: We can remove requestId from the CLI, its an optional parameter that is more handy for web applications, you can pass nil to CreateCallbackWithResponse for the params.

cmd/callbacks/create.go
```diff
@@ -0,0 +1,104 @@
...
+					cmd.Printf("Callback exists for the promise: %s\n", promiseId)
+				} else {
+					cmd.PrintErrln(res.Status(), string(res.Body))
+				}
+			}
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&promiseId, "promise-id", "", "promise id")
```
dfarr, line 98: We can mark this one as required

cmd/callbacks/create.go
```diff
@@ -0,0 +1,104 @@
...
+	cmd.Flags().StringVar(&rootPromiseId, "root-promise-id", "", "root promise id")
```
dfarr, line 99: We can also mark this one as required

cmd/callbacks/create.go
```diff
@@ -0,0 +1,104 @@
...
+	cmd.Flags().DurationVar(&timeout, "timeout", 0, "callback timeout")
```
dfarr, line 100: I think we can put a sensible default here, maybe 1m?

cmd/callbacks/create.go
```diff
@@ -0,0 +1,104 @@
...
+	cmd.Flags().StringVar(&recv, "recv", "", "receive object")
```
dfarr, line 101: We can put default (literally) as the default for recv

dfarr (changes requested):
So sorry for the delay in getting back to you on this PR. The `recv` flag is so tricky because it can be either a string or a json object in the form `{"type": "", data: {}}`.

I played around a little with this, what do you think about this approach?

```go
var recv v1.Recv

if json.Valid([]byte(recvStr)) {
	var recv0 v1.Recv0

	if err := json.Unmarshal([]byte(recvStr), &recv0); err != nil {
		return err
	}
	if err := recv.FromRecv0(recv0); err != nil {
		return err
	}
} else {
	if err := recv.FromRecv1(recvStr); err != nil {
		return err
	}
}
```

This unmarshals the provided value to a `v1.Recv0` if the value provided by the user is valid json, otherwise we assume it's a string. That way the user can provide one or the other, for example:

```
# string example
go run ./... callbacks create --promise-id foo --root-promise-id bar --recv default

# json example
go run ./... callbacks create --promise-id foo --root-promise-id bar --recv '{"type":"poll", "data":{"group":"default","id":"0"}}'
```

Madraceee:
I am not well. I will look into this next week. Is that fine?

dfarr:
> I am not well. I will look into this next week. Is that fine?

Of course, take your time and hope you get well soon!

dfarr:
Hey @Madraceee I can take over this PR if you'd like? Let me know your status :)

dfarr:
Closing in favour of #499

PR #448, opened 2024-11-01
Enqueue one task per root_promise_id

avillega:
Enqueue a single task per root_promise_id. This is necessary to prevent extra work happening for a single promise id.


avillega:
@dfarr I am still unsure how to add dst to this change, any pointers will be appreciated.

internal/app/subsystems/aio/store/store.go
```diff
@@ -71,3 +73,13 @@ func Collect(c <-chan *bus.SQE[t_aio.Submission, t_aio.Completion], f <-chan int
 
 	return batch, true
 }
+
+func StoreErr(err error) error {
```
dfarr, line 77: Damn, this looks really helpful!

dfarr:
Sorry for being anal here:
```
enqueable ->
enqueueable
```

internal/app/subsystems/aio/store/store.go
avillega: I hated that the errors had no context just, `expected 3 args got 4` or something like that. This at least points to the file and line where that happens.

dfarr (approved):
🎉

PR #470, opened 2024-11-20
Complete tasks based on root promise

avillega:


dfarr:
This looks good! I think the linearizability check is just being preemptively canceled before it has the chance to run because the lint failure, so hopefully if we fix the linter we should be good 🤞

dfarr (approved):

PR #478, opened 2024-12-09
Avillega/deduplicate callbacks

avillega:


internal/app/coroutines/createCallback.go
```diff
@@ -94,25 +97,30 @@ func CreateCallback(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any
...
+					Id:        callbackId,
+					PromiseId: r.CreateCallback.PromiseId,
+					Recv:      r.CreateCallback.Recv,
+					Mesg:      mesg,
+					Timeout:   r.CreateCallback.Timeout,
+					CreatedOn: createdOn,
 				}
-			} else {
-				return CreateCallback(c, r)
+				status = t_api.StatusCreated
```
dfarr: Should this be set in the if branch above and the else branch omitted?

dfarr (changes requested):

internal/app/coroutines/createCallback.go
avillega: Done!

dfarr (approved):

PR #485, opened 2024-12-12
Add id to callbacks in openapi spec

dfarr:


avillega:
Editing the title of the PR to say callbacks instead of tasks.

avillega (approved):

PR #495, opened 2024-12-14
[fix]: Resolve string receiver in sender subsystem (#467)

### LINKED ISSUE #467: Resolve string receiver in sender subsystem
A task recv value can be specified by the user in four locations:
1. as a tag on a promise (defaults to the `resonate:invoke` tag)
2. in a create callback request
3. in a create promise and callback request
4. in a create promise and task request

Of these four possible ways to specify a recv, only the first one goes through the router subsystem. In the router subsystem we have a convenience feature to resolve a recv specified as a string to an object if it is specified as a url that matches a predefined format. Because we do this matching in the router subsystem, the convenience only applies to the first location, but if we move the matching logic to the sender subsystem we can apply the convenience to all locations.

## Background

**Router Subsystem**

The router subsystem is responsible for creating a task for a promise based on some information on the promise, usually the `resonate:invoke` tag.

**Sender Subsystem**

The sender subsystem is responsible for sending a task message. For example, making an http request or sending a message to a message queue.

## TODO

We should be able to move the convenience matching function [here](https://github.com/resonatehq/resonate/blob/main/internal/app/subsystems/aio/router/router.go#L264-L293) to the sender subsystem, effectively applying the convenience function lazily.

TusharMohapatra07:
Fixes #467 

- I've added the `schemeToRecv` function to sender subsystem.
- Removed `schemeToRecv` from router subsystem.


TusharMohapatra07:
cc: @flossypurse @dfarr

dfarr (approved):
🎉

TusharMohapatra07:
@dfarr Thanks for the approval! 😁

PR #498, opened 2024-12-18
Log non linearizable path when linearization check fails

dfarr:
Thanks to anishathalye/porcupine#21 we can grab the longest linearizable path plus one additional operation (the first op that is not part of the linearizable path chronologically) to reproduce our DST logs including the error message. This information was lost in previous implementations of DST with porcupine.

When linearization fails, there may be multiple linearizable paths and multiple possible next operations. This implementation grabs the first non-linearizable path and the first possible next operation so the log is a "best effort" and should be taken with a grain of salt. As such the additional logs can be accessed behind the verbose flag.

```
go run ./... dst run -v
```

### DIFF
--- a/cmd/dst/run.go
+++ b/cmd/dst/run.go
@@ -32,6 +32,7 @@ func RunDSTCmd() *cobra.Command {
 		timeout           time.Duration
 		scenario          string
 		visualizationPath string
+		verbose           bool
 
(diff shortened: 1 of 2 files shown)


avillega (approved):
I really like that you also add the partition function, it will be useful when we do horizontal paritions.

PR #502, opened 2024-12-19
Create callback returns 403 when promise already completed

### LINKED ISSUE #500: Create callback returns 200 in case of non pending promise
The create callback request returns a 200 response code when registering a callback on a promise that is already completed. We should probably return a 403 by reusing our [StatusPromiseAlready*](https://github.com/resonatehq/resonate/blob/main/internal/kernel/t_api/status.go#L17-L20) statuses (we should be able to use [this](https://github.com/resonatehq/resonate/blob/main/internal/app/coroutines/completePromise.go#L213) helper function too).

DST and unit tests will need to be updated accordingly.

dfarr:
Resolves #500

### DIFF
--- a/test/dst/dst.go
+++ b/test/dst/dst.go
@@ -313,7 +313,7 @@ func (d *DST) Run(r *rand.Rand, api api.API, aio aio.AIO, system *system.System)
 func (d *DST) logNonLinearizable(ops []porcupine.Operation, history porcupine.LinearizationInfo) {
 	// log the linearizations
 	linearizations := history.PartialLinearizationsOperations()
-	util.Assert(len(linearizations) == 3, "linearizations must be equal to the number of partitions")
+	util.Assert(len(linearizations) == 4, "linearizations must be equal to the number of partitions")
 
 	// check each parition individually
 	// partitions are in order: promise, schedule, lock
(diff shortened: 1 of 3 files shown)


test/dst/dst.go
```diff
@@ -313,7 +313,7 @@ func (d *DST) Run(r *rand.Rand, api api.API, aio aio.AIO, system *system.System)
 func (d *DST) logNonLinearizable(ops []porcupine.Operation, history porcupine.LinearizationInfo) {
 	// log the linearizations
 	linearizations := history.PartialLinearizationsOperations()
-	util.Assert(len(linearizations) == 3, "linearizations must be equal to the number of partitions")
+	util.Assert(len(linearizations) == 4, "linearizations must be equal to the number of partitions")
```
dfarr, line 316: This just fixes a whoopsie that made it to main

avillega (approved):

avillega (commented):
If we try to create a callback for a promise that is already completed, shouldn't we return the promise as well? That way the sdks can avoid having to do an extra request for the promise.

dfarr:
I totally forgot we do this so that any process creating a callback can just continue when the promise complete, duh! Will close this PR.

I discovered this "issue" when using the new callback cli. Let's just log a different message indicating "the promise is already completed" in this case.

PR #522, opened 2025-01-09
Remove printing of a struct with a pointer in it

avillega:
This should fix the diff between postgress and sqlite runs in our latests dst runs.


avillega:
seems to correctly fix the problem. I run it with the failing seed and now is passing https://github.com/resonatehq/resonate/actions/runs/12700179355

dfarr (approved):

PR #531, opened 2025-01-15
Add notifications api

### LINKED ISSUE #513: Notify message
Add a notify message to our messages to support the bridge between the ephemeral and durable world.

A client can register a notify on a promise. Like registering a callback, registering a notify against a completed promise returns immediately with the promise. Registering a notify against a pending promise will result in a "best effort" message when the promise completes. In the event that the client is no longer available at this point in time the message will be discarded, the client is therefore responsible for re-registering for a notify message on restart (if still needed).

avillega:


avillega:
This is not fully ready yet, but most of the implementation is done, still need to write all the tests and make sure this fully works, but would be nice to get early feedback.

dfarr:
This looks great!

One thing that I'm not sure about atm is how many times we should (re)attempt an enqueue for a notify message? Our default logic is to retry until the timeout time (which is set to the same time as the promise timeout), does this make sense for notify? It's certainly the easiest to implement because we can just leave the logic as is.

avillega:
Our other option is to just try to send it and do not care about any errors. I think the spirit of notify goes more in that direction.

dfarr:
I'm up for that!

cmd/notify/create.go
```diff
@@ -0,0 +1,96 @@
...
+				Recv:      recv,
+			}
+
+			res, err := c.V1().CreateNotifyWithResponse(context.TODO(), nil, body)
+			if err != nil {
+				return err
+			}
+
+			if res.StatusCode() == 201 {
+				cmd.Printf("Created callback: %s\n", id)
```
dfarr:
```suggestion
				cmd.Printf("Created notification: %s\n", id)
```

cmd/notify/create.go
```diff
@@ -0,0 +1,96 @@
...
+			} else if res.StatusCode() == 200 {
+				if res.JSON200.Promise != nil && res.JSON200.Promise.State != v1.PromiseStatePENDING {
+					cmd.Printf("Promise %s already %s\n", promiseId, strings.ToLower(string(res.JSON200.Promise.State)))
+				} else {
+					cmd.Printf("Created callback: %s (deduplicated)\n", id)
```
dfarr:
```suggestion
					cmd.Printf("Created notification: %s (deduplicated)\n", id)
```

dfarr (approved):
Damn man, even the cli is here!

PR #537, opened 2025-01-24
Send to exact match only when notify in poller

dfarr:


internal/app/plugins/poll/poll.go
```diff
@@ -305,6 +306,11 @@ func (w *PollWorker) Process(mesg *aio.Message) {
 		return
 	}
 
+	if mesg.Type == message.Notify && conn.id != data.Id {
+		mesg.Done(false, fmt.Errorf("no connection found for group %s and id %s", data.Group, data.Id))
```
avillega, line 310: Given that not being able to deliver a notify message is not an error in that our guarantee is "at most once", should we return success here?

avillega (approved):

internal/app/plugins/poll/poll.go
dfarr: I was wondering the exact same thing!

The reason I went with an error is so that the sender subsystem _could_ retry notify messages if we ever want to do an enqueue retry loop (but still go directly to completed on successful enqueue). I figure the poller in the sdk may lose and re-establish a connection from time to time so it is possible that at one moment there is no active connection and later there is where the sdk instance itself did not actually restart.

PR #552, opened 2025-02-24
Add more space for errors...

avillega:
By making internal codes 1 digit larger, this should give us about an extra 90 error codes per http code.


avillega:
I know we talked about making the http error the last three digits, that will give us as bigger amount of errors to use, but it was ugly. I think 90 extra errors per http code should take us far.

dfarr (approved):

dfarr:
> I know we talked about making the http error the last three digits, that will give us as bigger amount of errors to use, but it was ugly. I think 90 extra errors per http code should take us far.

I agree, every code having the same number of digits is nice!

Issue #526, opened 2025-01-10
Horizontal partitioning for porocupine


avillega:
## Describe the problem you are facing
Our DST partitions are based on the types of our domain, Promises, Tasks, Schedules, Locks. This kind of partition doesn't scale well as we add more types that are related to the others, for example, Tasks need to have information about the Promise for some validations. 

## Describe the solution you'd like
Instead of having the partition by type, do partitions by id. The partitions can be done by root_promise_id and schedule_id. This will allows us to scale our DST and linearizability checks to run larger tests.

flossypurse:
Status update:

Bug found in the server, going to fix as part of this task.

Bug is "three invocation tasks" are created which shouldn't be possible.

Issue #514, opened 2025-01-09
Fix possibility of non atomic promise+task creation for CreatePromiseAndTask


avillega:
## Problem:
In some instances it is possible to end up with the creation of and claim of a task where the promise was not created. This happens because on conflict for promise creation we do nothing and continue as if the promise was created.

## Describe the solution you'd like
Make promise creation fail on conflict. This will require a small refactor given that we still need to ignore creation conflicts when dealing with promises created by schedules.

## Additional context
We found this behaviour when doing Callback+Promise creation. Given that this operation doesn't need to be atomic we split the two operations. However, we can do the same for Task+Promise creation given that this operation needs to be atomic.

PR #556, opened 2025-03-06
DST horizontal partitioning

avillega:
This PR contains several related changes to be able to make the DST run faster and enable validations between promises and tasks.

- Make DST do horizontal partitions based on the rootPromiseId
- Add many validations between tasks and promises as well as adjusting some others
- Make a complete promise operation also complete all related tasks including the current claimed task
- General small changes to the way we print and report DST results to make it easier to debug.
- Make trying to complete a task that is already completed return a StatusOk response
- Create a single CreatePromiseAndTask command that is aware of not creating a task in case a promise was not being able to be created.

### DIFF
--- a/internal/app/coroutines/createCallback.go
+++ b/internal/app/coroutines/createCallback.go
@@ -66,7 +66,7 @@ func CreateCallback(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any
 
 			createdOn := c.Time()
 
-			callbackId := fmt.Sprintf("%s.%s", r.CreateCallback.PromiseId, r.CreateCallback.Id)
+			cbId := callbackId(r.CreateCallback.RootPromiseId, r.CreateCallback.PromiseId)
 			completion, err := gocoro.YieldAndAwait(c, &t_aio.Submission{
 				Kind: t_aio.Store,
 				Tags: r.Tags,
@@ -76,7 +76,7 @@ func CreateCallback(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any
 							{
 								Kind: t_aio.CreateCallback,
 								CreateCallback: &t_aio.CreateCallbackCommand{
-									Id:        callbackId,
+									Id:        cbId,
 									PromiseId: r.CreateCallback.PromiseId,
 									Recv:      r.CreateCallback.Recv,
 									Mesg:      mesg,
@@ -104,7 +104,7 @@ func CreateCallback(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any
 			if result.RowsAffected == 1 {
 				status = t_api.StatusCreated
 				cb = &callback.Callback{
-					Id:        callbackId,
+					Id:        cbId,
 					PromiseId: r.CreateCallback.PromiseId,
 					Recv:      r.CreateCallback.Recv,
 					Mesg:      mesg,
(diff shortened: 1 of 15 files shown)


internal/app/coroutines/createPromise.go
```diff
@@ -281,8 +307,21 @@ func createPromise(tags map[string]string, promiseCmd *t_aio.CreatePromiseComman
...
+		if isCreatePromiseAndTask {
+			promiseAndTaskResult := completion.Store.Results[0].CreatePromiseAndTask
+			util.Assert(promiseAndTaskResult.PromiseRowsAffected == 0 || promiseAndTaskResult.PromiseRowsAffected == 1, "Creating promise result must return 0 or 1 rows")
+			if promiseAndTaskResult.PromiseRowsAffected == 0 {
+				util.Assert(promiseAndTaskResult.TaskRowsAffected == 0, "If not promise was created a task must have not been created")
+			}
+		} else {
+			createPromiseResult := completion.Store.Results[0].CreatePromise
+			util.Assert(createPromiseResult.RowsAffected == 0 || createPromiseResult.RowsAffected == 1, "CreatePromise result must return 0 or 1 rows")
+		}
```
avillega, line 319: As a general note, I might want to come back to this code and refactor it in a cleaner way. the `isCreatePromiseAndTask` every where makes me believe there is a cleaner approach to it.

internal/app/coroutines/createCallback.go
```diff
@@ -138,3 +138,7 @@ func CreateCallback(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any
 	util.Assert(res != nil, "response must not be nil")
 	return res, nil
 }
+
+func callbackId(rootPromiseId, promiseId string) string {
+	return fmt.Sprintf("__resume:%s:%s", rootPromiseId, promiseId)
```
dfarr, line 143: I would be up for just `resume:...` and omit the underscores, what do you think?

test/dst/dst.go
```diff
@@ -245,16 +214,26 @@ func (d *DST) Run(r *rand.Rand, api api.API, aio aio.AIO, system *system.System)
 				// our model has been updated via the backchannel
 				counter := obj.Counter - r.Intn(2)
 
+				// The ProcessId is always the taskId, which means each task
+				// is always claimed by a different process, which means
+				// that when hearthbeating there will always be a single task at
```
dfarr:
```suggestion
				// that when heartbeating there will always be a single task at
```

test/dst/validator.go
```diff
@@ -655,12 +681,33 @@ func (v *Validator) ValidateCompleteTask(model *Model, reqTime int64, resTime in
...
 		model.tasks.set(req.CompleteTask.Id, res.CompleteTask.Task)
 		return model, nil
-	case t_api.StatusTaskAlreadyCompleted:
+	case t_api.StatusOK:
 		if t == nil {
 			return model, fmt.Errorf("task '%s' does not exist", req.CompleteTask.Id)
 		}
 		if !t.State.In(task.Completed|task.Timedout) && t.Timeout >= resTime {
-			return model, fmt.Errorf("task '%s' state not completed", req.CompleteTask.Id)
+			// This could happen if the promise timetout
```
dfarr:
```suggestion
			// This could happen if the promise timedout
```

dfarr (approved):
Very nice! This ended up being quite a chunk of work

internal/app/coroutines/createCallback.go
dfarr: (and the same for the others as well)

avillega: I am usually very paranoiac about name collisions in these cases that is why I like to add the double underscore. I think we don' lose much by having the underscores but might increase our risk of collisions or confusion without it. but I am also not strongly on necessarily having them.

avillega:
This PR fixes #526 and #514

PR #583, opened 2025-03-18
Reorganize the dst html tooltip table

dfarr:


test/dst/dst.go
```diff
@@ -495,70 +495,59 @@ func (d *DST) Model() porcupine.Model {
...
-					            </tr>
-					          </thead>
-					          <tbody>
-					            %s
-					          </tbody>
-					        </table>
-					      </td>
-					    </tr>
-					  </tbody>
+						<!-- First Row: Promises & Tasks -->
```
avillega, line 498: Should be just promises

avillega (approved):

test/dst/dst.go
avillega:
```suggestion
						<!-- First Row: Promises-->
```

PR #584, opened 2025-03-18
DST Fix

avillega:
Fixes most failures so far.

### DIFF
--- a/internal/app/coroutines/completePromise.go
+++ b/internal/app/coroutines/completePromise.go
@@ -166,7 +166,7 @@ func completePromise(tags map[string]string, cmd *t_aio.UpdatePromiseCommand, ad
 				Kind: t_aio.CreateTasks,
 				CreateTasks: &t_aio.CreateTasksCommand{
 					PromiseId: cmd.Id,
-					CreatedOn: cmd.CompletedOn,
+					CreatedOn: c.Time(),
 				},
 			},
 			{
--- a/internal/app/coroutines/completePromise.go
+++ b/internal/app/coroutines/completePromise.go
@@ -159,7 +159,7 @@ func completePromise(tags map[string]string, cmd *t_aio.UpdatePromiseCommand, ad
 				Kind: t_aio.CompleteTasks,
 				CompleteTasks: &t_aio.CompleteTasksCommand{
 					RootPromiseId: cmd.Id,
-					CompletedOn:   cmd.CompletedOn,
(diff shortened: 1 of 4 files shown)


test/dst/dst.go
```diff
@@ -490,8 +497,9 @@ func (d *DST) Model() porcupine.Model {
 						<td align="right">%d</td>
 						<td align="right">%d</td>
 						<td align="right">%d</td>
+						<td align="right">%s</td>
 					</tr>
-				`, t.value.Id, t.value.State, t.value.RootPromiseId, t.value.ExpiresAt, t.value.Timeout, *t.value.CreatedOn)
+				`, t.value.Id, t.value.State, t.value.RootPromiseId, t.value.ExpiresAt, t.value.Timeout, *t.value.CreatedOn, completedOn)
```
dfarr, line 502: Not super important but we have a utility function for this purpose
```suggestion
				`, t.value.Id, t.value.State, t.value.RootPromiseId, t.value.ExpiresAt, t.value.Timeout, *t.value.CreatedOn, util.SafeDeref(t.value.CompletedOn))
```

I also think the indenting is off a little on this line.

internal/app/coroutines/completePromise.go
```diff
@@ -159,14 +159,14 @@ func completePromise(tags map[string]string, cmd *t_aio.UpdatePromiseCommand, ad
 				Kind: t_aio.CompleteTasks,
 				CompleteTasks: &t_aio.CompleteTasksCommand{
 					RootPromiseId: cmd.Id,
-					CompletedOn:   cmd.CompletedOn,
```
dfarr, line 162: Is it possible to remove the CompletedOn key from CompleteTasksCommand now?

dfarr (changes requested):

test/dst/dst.go
avillega: Yes, it will give me 0 instead of "--" which is what I wanted, might make the code simplier though

internal/app/coroutines/completePromise.go
avillega: Also remove it from the tasks table? I think it might be useful and setting it here gives us flexibility and some good information when debugging.

dfarr: Oh whoops, I was thinking this was a request but it's a command so we do need completedon

PR #681, opened 2025-05-14
Merge create callback and create subscrition

dfarr:
These will be separate api subsystem requests but the same coroutine.

I intended to change the routes to:
```
/promises/{id}/callback
/promises/{id}/subscribe
```

However we use a catch-all for the id parameter, which means the id may include a "/" and additional parts cannot be matched following this parameter. I have left the http routes as-is.


internal/app/coroutines/createCallback.go
```diff
@@ -1,22 +1,25 @@
...
-	var res *t_api.Response
-
-	if r.CreateCallback.PromiseId == r.CreateCallback.RootPromiseId {
+	util.Assert(r.Kind == t_api.CreateCallback, "kind must be create callback")
+	util.Assert(r.CreateCallback != nil, "create callback must not be nil")
+	util.Assert(r.CreateCallback.Mesg.Type == "resume" || r.CreateCallback.Mesg.Type == "notify", "message type must be resume or notify")
+	util.Assert(r.CreateCallback.Mesg.Type == "resume" || r.CreateCallback.PromiseId == r.CreateCallback.Mesg.Root, "if notify, root promise id must equal leaf promise id")
+	// util.Assert(r.CreateCallback.Mesg.Type == "notify" ||(r.CreateCallback.PromiseId == r.CreateCallback.Mesg.Leaf && r.CreateCallback.PromiseId != r.CreateCallback.Mesg.Root), "if resume, root promise id must not equal leaf promise id")
+
+	// TODO: should this be a validation in the api layer?
```
dfarr: @avillega how do you feel about moving this to the api layer?

avillega: I think I prefer it in the coroutine. It is kind of a business logic validation, and it applies to all api subsystems. we would have to do the validation in the http subsystem and grpc subsystem and any future ones. I prefer when all the validations are close together so it is easy to track where stuff fails. Also, if it is in the coroutine it can be tested by DST, if we put it in the api layer we won't test it.

avillega (approved):
Don't forget to update the openapi.yml file. I am unsure if this could break the SDK given that the createcallback request does not accept an id anymore.

dfarr:
> Don't forget to update the openapi.yml file. I am unsure if this could break the SDK given that the createcallback request does not accept an id anymore.

Thanks for the reminder! I verified you can still include id and our server just ignore it so we won't break the sdk (also the tests still include the id parameter, however I should probably remove that too)

internal/app/coroutines/createCallback.go
dfarr: I agree that it would be preferable to centralize the validation. The tricky thing is I would also prefer to "fail fast". 

Hmmm okay what about building out a validation mechanism in the kernel api implemented [here](https://github.com/resonatehq/resonate/blob/main/internal/api/api.go#L136)? I imagine we could have hooks based on the request kind, and the best part is, this is the interface that [dst uses](https://github.com/resonatehq/resonate/blob/main/test/dst/dst.go#L152) too!

dfarr:
I have updated the open api spec and added api validators. I really like the api validators and suspect there are other things we could use them for, I figure we can do this in a step by step approach though.

test/dst/dst.go
```diff
@@ -658,6 +655,13 @@ func (d *DST) Step(model *Model, reqTime int64, resTime int64, req *t_api.Reques
 			return model, nil
 		case t_api.StatusSchedulerQueueFull:
 			return model, nil
+		case t_api.StatusFieldValidationError:
+			if req.Kind == t_api.CreateCallback && req.CreateCallback.Mesg.Type == "resume" && req.CreateCallback.PromiseId == req.CreateCallback.Mesg.Root {
+				// sometimes we generate create callback requests with the same root and
+				// leaf promise ids by chance
+				return model, nil
+			}
+			fallthrough
```
dfarr, line 665: I am both so happy and so horrified that I got to use a `fallthrough`

PR #683, opened 2025-05-16
Poller support for uni and any

dfarr:
```
# unicast
poll://uni@default/1

# anycast
poll://any@default

# anycast with preference
poll://any@default/2
```


avillega (approved):
As expected no good way to use any and uni cast for the http sender system.

dfarr:
> As expected no good way to use any and uni cast for the http sender system.

Not at all, it trips me up so much. It seems to me to be an implicit part of the address you specify, for example:
- url that points to a load balancer with (optional) sticky routing: anycast
- url that resolves to a single node: unicast

PR #691, opened 2025-05-21
Refactor our usage of sum-types

avillega:
In general this PR changes the following:
- Changes the definition of `t_api.Request` to be closer to the recommended way of defining sum types in Go. Our current implementation stores many pointers per request when only one is needed. This PR improves on that.
- Change `Request.Tags` to `Request.Metadata` in order to reduce confusion and term overloading with Promise Tags.
- Move the request validation to a method of the RequestPayload.


internal/app/coroutines/acquireLock.go
```diff
@@ -11,22 +11,24 @@ import (
 )
 
 func AcquireLock(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any], r *t_api.Request) (*t_api.Response, error) {
-	expiresAt := c.Time() + r.AcquireLock.Ttl
+	acquireLockReq := r.Payload.(*t_api.AcquireLockRequest)
```
dfarr: This is really nit picky so I apologize in advance.

How would you feel about naming this variable (and all similar ones) just req? The context is defined by the function/file and once I'm in that context I find simpler variable names help reduce cognitive overhead.

internal/kernel/t_api/request.go
```diff
@@ -10,39 +11,16 @@ import (
...
-	// TASKS
-	ClaimTask      *ClaimTaskRequest
-	CompleteTask   *CompleteTaskRequest
-	DropTask       *DropTaskRequest
-	HeartbeatTasks *HeartbeatTasksRequest
+type RequestPayload interface {
+	Kind() Kind
+	String() string
+	Validate() error
+	isRequestPayload()
```
dfarr, line 18: What is this function used for? I wasn't able to find any usages but may have overlooked one

internal/kernel/t_api/request.go
```diff
@@ -10,39 +11,16 @@ import (
...
+}
 
-	// ECHO
-	Echo *EchoRequest
+type Request struct {
```
dfarr, line 21: Sorry about the nitpicky naming stuff again, how would you feel about naming these fields `Head` and `Body`?

internal/kernel/t_api/request.go
```diff
@@ -115,6 +143,19 @@ func (r *CreateCallbackRequest) String() string {
 	return fmt.Sprintf("CreateCallback(id=%s, promiseId=%s, recv=%s, mesg=%s, timeout=%d)", r.Id, r.PromiseId, r.Recv, r.Mesg, r.Timeout)
 }
 
+func (r *CreateCallbackRequest) Validate() error {
```
dfarr, line 136: alright this is pretty awesome!

dfarr:
Okay this is really awesome, I feel dumb, why would we do it any other way in the first place!!!???

internal/kernel/t_api/request.go
avillega: This is the empty method that signals that a struct is part of this sum type. There should be no usages of it, but any new type that needs to be a requestPayload needs to provide an empty implementation of it.

internal/app/coroutines/acquireLock.go
avillega: I like it.

internal/kernel/t_api/request.go
avillega: I think I prefer Metadata rather to Head. The plan after this PR is to change the other types to follow this same pattern, Head seems specific to Requests, but we add the Metadata to all the commands of the system.

internal/kernel/t_api/request.go
avillega: I can probably put all the `isRequestPayload` implementations together so it is evident that it is a marker method and not to attach functionality to it.

dfarr: Oh gotchya, is it sufficient that a struct implements the rest of the interface? kind / string / validate?

avillega: if a struct implements the rest of the interface and not that one it will not be a requestPayload and it will not be possible to use it where a requestPayload is expected. is part of the type safety of the feature.

PR #692, opened 2025-05-23
Refactor Response sum type

avillega:
Same deal as with the Request type

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -154,7 +154,7 @@ func (a *api) EnqueueSQE(sqe *bus.SQE[t_api.Request, t_api.Response]) {
 
 			status = error.Code()
 		} else {
-			status = res.Status()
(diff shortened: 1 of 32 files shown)


dfarr (approved):

dfarr:
Are we doing `.As**()` type assertions for requests too?

avillega:
> Are we doing `.As**()` type assertions for requests too?

I didn't do it for Requests, I guess I didn't need it as much. However I think I will eventually add it just for consistency and because is a nice pattern.

dfarr:
> > Are we doing `.As**()` type assertions for requests too?
> 
> I didn't do it for Requests, I guess I didn't need it as much. However I think I will eventually add it just for consistency and because is a nice pattern.

Yeah if for no other reason it would be nice for the consistency

PR #702, opened 2025-05-26
Refactor Store Sum Types

avillega:
Similar deal as with the Request/Response types. The main difference being that it was possible to just have the type instead of having to wrap it in a "Payload" given that this types do not hold metadata.


dfarr (approved):
Net ~2000 lines of code removed! That's awesome 🎉

PR #787, opened 2025-09-03
Implement resonate dev alias

### LINKED ISSUE #746: Add dev command
```
resonate dev
```

or

```
resonate serve --dev
```

This command is an alias for `resonate serve --aio-store-sqlite-path :memory:`. We can use it in our documentation to make initial development simpler, restarting the sever clears unintended state the database may end up in during development.

avillega:
Resolves #746

### DIFF
--- a/cmd/dev/dev.go
+++ b/cmd/dev/dev.go
@@ -13,7 +13,7 @@ func DevCmd() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "dev",
-		Short: "Start Resonate in development mode with in-memory SQLite store",
+		Short: "Start Resonate server in development mode",
 		Long:  "Start Resonate server with development-friendly defaults (in-memory SQLite store).\n\nThis command is an alias for: resonate serve --aio-store-sqlite-path :memory:",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := config.Parse(); err != nil {
--- a/cmd/dev/dev.go
+++ b/cmd/dev/dev.go
@@ -1,0 +1,37 @@
+package dev
+
+import (
+	"github.com/resonatehq/resonate/cmd/config"
+	"github.com/resonatehq/resonate/cmd/serve"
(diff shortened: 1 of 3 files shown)


cmd/serve/serve.go
```diff
@@ -173,7 +161,22 @@ func ServeCmd() *cobra.Command {
...
+}
+
+func ServeCmd() *cobra.Command {
+	var (
+		config = &config.Config{}
+	)
+
+	cmd := &cobra.Command{
+		Use:   "serve",
+		Short: "Resonate server",
```
dfarr:
```suggestion
		Short: "Start Resonate server",
```

cmd/dev/dev.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func DevCmd() *cobra.Command {
+	var (
+		config = &config.Config{}
+	)
+
+	cmd := &cobra.Command{
+		Use:   "dev",
+		Short: "Start Resonate in development mode with in-memory SQLite store",
```
dfarr:
```suggestion
		Short: "Start Resonate server in development mode",
```

cmd/dev/dev.go
```diff
@@ -0,0 +1,37 @@
...
+		Long:  "Start Resonate server with development-friendly defaults (in-memory SQLite store).\n\nThis command is an alias for: resonate serve --aio-store-sqlite-path :memory:",
```
dfarr, line 17: I wonder if we should remove this because we could extend the dev config, could go either way on this one

dfarr (changes requested):

cmd/dev/dev.go
avillega: Lets keep it on the long description, once we extend the dev config we can figure out the best help message, I think for know it clearly states what the command does and the expectations that it run an in memory db so no persistance is expected

dfarr (approved):

PR #788, opened 2025-09-03
Add windows binaries to our release artifacts

avillega:


avillega:
To test this one, we can crate a test tag and see if it creates the right artifacts for it and all the OSs.

dfarr (approved):

PR #812, opened 2025-09-15
Convert task ttl to bigint

dfarr:
The max possible ttl is now 9223372036854775807ms or ~292471208.677536 years.

On heartbeat, we calculate the sum of the current time in ms and the ttl value. Because this calculation happens in the database query, there is no good way to clamp this value to avoid overflow. I propose we set reasonable maximums for the following values:
- the current time in ms
- the ttl

Setting a bound on the current time will actually be pretty easy because we can do it in assert at the top of the tick. We can choose a value so ridiculously large that the clock should never hit it even in a simulation. If this assertion is triggered, it is likely somebody is messing with the machines clock.

Then we would just need to choose a max ttl value such that:
```
max_clock + ttl <= 9223372036854775807
```

But let's create an issue for this and do it later.


internal/app/subsystems/aio/store/postgres/postgres.go
```diff
@@ -101,7 +101,7 @@ const (
 		timeout         BIGINT,
 		counter         INTEGER DEFAULT 1,
 		attempt         INTEGER DEFAULT 0,
-		ttl             INTEGER DEFAULT 0,
+		ttl             BIGINT DEFAULT 0,
```
avillega, line 104: Do you know if we need to change this in sqlite?

internal/kernel/t_api/request.go
```diff
@@ -299,7 +299,7 @@ func (r *HeartbeatLocksRequest) Kind() Kind {
 type CreateTaskRequest struct {
 	PromiseId string `json:"promiseId"`
 	ProcessId string `json:"processId"`
-	Ttl       int    `json:"ttl" binding:"min=0"`
+	Ttl       int64  `json:"ttl" binding:"min=0"`
```
avillega, line 302: does the request fails if the json has a number out of range? should we add a test for it?

avillega (approved):

internal/app/subsystems/aio/store/postgres/postgres.go
dfarr: Luckily no, I looked into it and sqlite INTEGER is dynamic

internal/kernel/t_api/request.go
dfarr: Yes it does and it results in a 500 response code, although a 400 would be appropriate. Let me create an issue for this, if we can "catch" these specific errors and map them to 400 response codes it would be good to implement that for all our integer fields.

dfarr: Nevermind, we already have the correct response code! Let me see if I can add a quick test:
```
❯ curl -i --json '{"id": "foo", "timeout": 9223372036854775808}' http://localhost:8001/promises
HTTP/1.1 400 Bad Request
Content-Type: application/json; charset=utf-8
Date: Tue, 16 Sep 2025 18:27:06 GMT
Content-Length: 315

{"error":{"code":40000,"message":"The request is invalid","details":[{"@type":"FieldValidationError","message":"json: cannot unmarshal number 9223372036854775808 into Go struct field createPromiseBody.timeout of type int64","domain":"request","metadata":{"url":"https://docs.resonatehq.io/operate/errors#40000"}}]}}%
```

PR #813, opened 2025-09-17
Add fencing token

avillega:


avillega:
I think this is almost completed, the hardest part is testing, but we can get the review started.

cmd/serve/serve.go
avillega, line 1: This file was unformatted, this changes are format only

internal/kernel/t_api/status.go
```diff
@@ -23,6 +23,7 @@ const (
 	StatusTaskAlreadyCompleted   StatusCode = 40306
 	StatusTaskInvalidCounter     StatusCode = 40307
 	StatusTaskInvalidState       StatusCode = 40308
+	StatusFencingTokenInvalid    StatusCode = 40309
```
dfarr: Can we use code 41200, which would map to a [412 precondition failed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/412)

internal/kernel/t_aio/store.go
```diff
@@ -20,7 +20,10 @@ func (s *StoreSubmission) String() string {
 }
 
 type Transaction struct {
-	Commands []Command
+	// Use as the fencing token
+	TaskId      string
```
dfarr: Can we wrap `TaskId` and `TaskCounter` into a struct and pass a pointer? We can reserve nil for no precondition.

internal/kernel/t_api/status.go
```diff
@@ -68,6 +69,8 @@ func (s StatusCode) String() string {
 		return "The task counter is invalid"
 	case StatusTaskInvalidState:
 		return "The task state is invalid"
+	case StatusFencingTokenInvalid:
+		return "The fencing token is invalid"
```
```suggestion
		return "The task precondition failed"
```

internal/kernel/t_api/status.go
```suggestion
	StatusTaskPreconditionFailed StatusCode = 41200
```

internal/app/coroutines/completePromise.go
```diff
@@ -32,6 +35,10 @@ func CompletePromise(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, an
 		return nil, t_api.NewError(t_api.StatusAIOStoreError, err)
 	}
 
+	if !completion.Store.Valid {
+		return nil, t_api.NewError(t_api.StatusFencingTokenInvalid, errors.New("Invalid task for requested operation"))
```
dfarr:
```suggestion
		return nil, t_api.NewError(t_api.StatusFencingTokenInvalid, errors.New("the specified task is not valid"))
```

dfarr (changes requested):
This looks seriously GREAT! My requests are mostly in regards to the way we communicate this error.

internal/kernel/t_aio/store.go
```diff
@@ -338,11 +339,16 @@ func (*HeartbeatLocksCommand) isCommand()       {}
 func (*TimeoutLocksCommand) isCommand()         {}
 
 type StoreCompletion struct {
+	Valid   bool
 	Results []Result
 }
 
 func (c *StoreCompletion) String() string {
-	return fmt.Sprintf("Store(results=%s)", c.Results)
+	if c.Valid {
```
dfarr: Could we just do this?

```go
return fmt.Sprintf("Store(valid=%t, results=%s)", c.Valid, c.Results)
```

dfarr (approved):
This looks really good! Let's get this in and do a release tomorrow :)

internal/kernel/t_aio/store.go
```diff
@@ -338,11 +339,16 @@ func (*HeartbeatLocksCommand) isCommand()       {}
...
+		return fmt.Sprintf("Store(results=%s)", c.Results)
+	} else {
+		return "Store(valid=false)"
+	}
```
avillega:
```suggestion
    return fmt.Sprintf("Store(valid=%t, results=%s)", c.Valid, c.Results)
```

PR #815, opened 2025-09-17
Plugin retry config

dfarr:
Adds the following plugin configs:
1. time to retry: the amount of time to wait before attempting to resend a task after a failed send attempt
2. time to claim: the amount of time wait until a task is claimed before attempting to resend a task after a successful send

This effectively moves the configuration to the plugin (it was previously global) and differentiates the two different times (previously treated the same). Also adds a special semantic for `State=Enqueued AND ExpiresAt =0`, meaning "never attempt to resend".


dfarr:
This PR also fixed a regression. Previously we waited for the ExpiresAt time to elapse before attempting to send, this helps avoid relentless sending. This feature has been re-added to the enqueue tasks coroutine.

internal/app/plugins/poll/poll.go
```diff
@@ -308,23 +312,42 @@ func (w *PollWorker) Process(mesg *aio.Message) {
...
 		return
 	}
 
 	// check if we have a connection
 	conn, ok := w.connections.get(addr)
 	if !ok {
-		mesg.Done(false, fmt.Errorf("no %s connection found for group %s and id %s", addr.Cast, addr.Group, addr.Id))
+		mesg.Done(&t_aio.SenderCompletion{
+			Success:     false,
+			Error:       fmt.Errorf("no %s connection found for group '%s' and id '%s'", addr.Cast, addr.Group, addr.Id),
```
avillega: I was thinking if we should make this error something like `no connection found for "poll://{group}/{id}"` since that is the format people will probably set in the sdks too and might be easier to understand

avillega (approved):

internal/app/plugins/poll/poll.go
dfarr: Nice, I like that will add

PR #844, opened 2025-10-03
Add resonate:delay tag

### LINKED ISSUE #843: Add resonate:delay

Currently, when we "resonate invoke", we cannot delay the invocation into the future

Add a "resonate:delay" tag (and parameters to rpc) to delay a function until the specified timestamp

## Additional context

I believe we can achieve this by delaying the task creation. Or we create the task but add a delay to the task

dfarr:
When this tag is set we delay sending a task until the time specified has elapsed. The time must be specified as an epoch value in ms.

example
```
resonate:delay=1767139200000  # delay until 2026-01-01T00:00:00Z
```

### DIFF
--- a/cmd/invoke/invoke.go
+++ b/cmd/invoke/invoke.go
@@ -36,17 +36,18 @@ func NewCmd() *cobra.Command {
 		funcName string
 		args     []string
 		jsonArgs string
+		version  int
 		timeout  time.Duration
 		target   string
+		delay    time.Duration
 		server   string
 		username string
 		password string
-		version  int
 	)
 
 	cmd := &cobra.Command{
 		Use:     "invoke <promise-id> --func <function-name> [flags]",
-		Short:   "Invoke a function (--func is required)",
+		Short:   "Invoke a function",
 		Example: invokeExample,
 		PreRunE: func(cmd *cobra.Command, cmdArgs []string) error {
 			if username != "" || password != "" {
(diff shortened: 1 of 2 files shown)


dfarr:
Resolves #843

cmd/invoke/invoke.go
```diff
@@ -120,19 +121,27 @@ func NewCmd() *cobra.Command {
...
 		},
 	}
 
 	cmd.Flags().StringVar(&funcName, "func", "", "function to invoke")
 	cmd.Flags().StringArrayVarP(&args, "arg", "a", []string{}, "function argument, can be provided multiple times")
 	cmd.Flags().StringVar(&jsonArgs, "json-args", "", "function arguments as json array")
+	cmd.Flags().IntVar(&version, "version", 1, "function version")
 	cmd.Flags().DurationVar(&timeout, "timeout", time.Hour, "promise timeout")
 	cmd.Flags().StringVar(&target, "target", "poll://any@default", "invoke target")
+	cmd.Flags().DurationVar(&delay, "delay", 0, "promise delay")
```
avillega, line 141: We should describe what is the effect of this flag a bit more, like mentioning that the promise is created but it only delays the invocation.

avillega (approved):

cmd/invoke/invoke.go
dfarr: Will follow up on this

PR #846, opened 2025-10-04
Sender plugin: Nats

azghr:
feat(plugin): add prototype NATS.io plugin for sender subsystem

- Add experimental NATS plugin with full NATS.io integration
- Support NATS URL parsing in sender subsystem (nats://host:port/subject)
- Implement physical receiver support for JSON format
- Add configurable workers, timeouts, and retry logic
- Include comprehensive test suite with mock clients
- Support various NATS URL formats (auth, multiple servers, wildcards)
- Integrate with existing router subsystem for tag-based matching
- Add metrics integration and error handling
- Maintain consistent snake_case naming conventions

Key new changes include:
- New `cmd/plugin_nats.go` to register the NATS plugin.
- Updates to `go.mod` to include NATS dependencies.
- Creation of new modules for the NATS plugin (`internal/app/plugins/nats`)
  and the NATS API subsystem (`internal/app/subsystems/api/nats`).
- Implementation of NATS request/response handling for various API
  operations (promises, schedules, locks, tasks).
- Updates to the sender subsystem to support the `nats://` scheme.


internal/app/plugins/nats/nats.go
```diff
@@ -0,0 +1,201 @@
...
+	client := w.client
+	if client == nil {
+		opts := []natsgo.Option{
+			natsgo.Timeout(w.timeout),
+			natsgo.RetryOnFailedConnect(false), // Disable retry for timeout testing
+			natsgo.PingInterval(20 * time.Second),
+			natsgo.MaxPingsOutstanding(2),
+		}
+
+		nc, err := natsgo.Connect(addr.URL, opts...)
```
dfarr: Can we move this to the worker construction and return an error immediately if we cannot connect?

internal/app/plugins/nats/nats.go
```diff
@@ -0,0 +1,201 @@
...
+	if w.client != nil {
+		err := client.Publish(addr.Subject, body)
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+
+	// Only use timeout logic for real NATS connections
+	ctx, cancel := context.WithTimeout(context.Background(), w.timeout)
```
dfarr: I think if we instantiate the client when we create a worker, we can always take this code path to take timeouts into account and remove the conditional call to publish above?

dfarr (changes requested):
Thank you so much @muhammad-asghar-ali you are on fire!

dfarr:
Sorry for the delay in getting back to you @muhammad-asghar-ali. I finally got around to testing this out (and learning more about nats). 

I noticed that we can specify the nats server host in the address, but we don't use it because the connection is created once at server startup and therefore we need to know the url of the nats server at this time.

I think this is a fine, even a preferable, setup. But it does make the host part of the uri misleading. I wonder if for the time being we only specify uris like `nats:///my-subject`? This seems to be a valid uri (I've seen it used for `file:///path` before) and it removes the misleading part, what do you think?

azghr:
> I noticed that we can specify the nats server host in the address, but we don't use it because the connection is created once at server startup and therefore we need to know the url of the nats server at this time.
> 
> I think this is a fine, even a preferable, setup. But it does make the host part of the uri misleading. I wonder if for the time being we only specify uris like `nats:///my-subject`? This seems to be a valid uri (I've seen it used for `file:///path` before) and it removes the misleading part, what do you think?

Thanks for reviewing, I'll look into your suggestion and share the solution with you

azghr:
> Sorry for the delay in getting back to you @muhammad-asghar-ali. I finally got around to testing this out (and learning more about nats).
> 
> I noticed that we can specify the nats server host in the address, but we don't use it because the connection is created once at server startup and therefore we need to know the url of the nats server at this time.
> 
> I think this is a fine, even a preferable, setup. But it does make the host part of the uri misleading. I wonder if for the time being we only specify uris like `nats:///my-subject`? This seems to be a valid uri (I've seen it used for `file:///path` before) and it removes the misleading part, what do you think?

You're right, `nats:///my-subject` is cleaner and removes the misleading host part.
Good news is that, `strings.TrimPrefix(u.Path, "/")` already works for both formats, so this is a zero-breaking-change improvement.
I'll update the test cases to use the new format and we're done.

dfarr:
Hey @muhammad-asghar-ali, we have decided to hold on off on additional plugins until we have a plugin architecture in place (see #886). Let's leave this PR as is for now and we will let you know when the new architecture is ready and we can then port this PR to the new design.

azghr:
Hey @dfarr need some guidance on github workflow failing, the error says 

```
Error: internal/app/plugins/nats/nats.go:112:59: cannot use proc (variable of type *processor) as base.Processor value in argument to base.NewPlugin: *processor does not implement base.Processor (wrong type for method Process)
		have Process([]byte, []byte) (bool, error)
		want Process([]byte, map[string]string, []byte) (bool, error)
Error: Process completed with exit code 1.
```
Basically, Go is complaining that my `processor` type doesn’t match the `base.Processor` interface. The method signature for Process is different: my implementation currently takes `([]byte, []byte)`, but the interface expects `([]byte, map[string]string, []byte)`.

So the solution is to update the Process method on `processor` to match the interface. Something like this mock solution could work:

```
// Old method
func (p *processor) Process(key []byte, value []byte) (bool, error) {
    // existing logic
    return true, nil
}

// Updated method to satisfy base.Processor
func (p *processor) Process(key []byte, headers map[string]string, value []byte) (bool, error) {
    // handle headers if needed
    return true, nil
}
```

Once the method signature matches, `processor` should correctly implement `base.Processor`, and the workflow error should disappear and also change the other plugin according to new signature. Need your thought on this before implementing.

dfarr:
Hey @muhammad-asghar-ali I think you could rebase this PR on the current head of main and then you should see the compilation errors locally. If you can fix those and push we should be good to go!

azghr:
> compilation

I merge the main with this feature branch `nats-plugin` and the build the run the resonate project locally, it working fine for... here is screen-short of steps I followed. Please guide me if I'm missing something!! @dfarr 

Screenshot 2025-11-05 at 12 58 50 AM

flossypurse:
Hi @azghr — just doing a maintenance pass on open PRs. Could you give us a quick status update on this one as well? Is it still in progress, or would you like us to close it for now? Thanks!

azghr:
> Hi @azghr — just doing a maintenance pass on open PRs. Could you give us a quick status update on this one as well? Is it still in progress, or would you like us to close it for now? Thanks!

Hi @flossypurse this feature is completed as per the new changes... waiting for you guys to review. Thanks!

dfarr:
Thank you once again for you contribution @azghr! We will soon be deprecating this repo in favour of a new server project (this repo will remain under a new name). In preparation for that I am closing all outstanding PRs.

PR #852, opened 2025-10-10
WIP Migrations

avillega:


avillega:
This is very WIP

internal/app/subsystems/aio/store/migrations/migrations.go
```diff
@@ -0,0 +1,198 @@
...
+	} else if regexp.MustCompile(`(?i)already exists|duplicate`).MatchString(errStr) {
+		suggestions += "- The object may already exist from a previous migration\n"
+		suggestions += "- Use IF NOT EXISTS clauses where appropriate\n"
+	} else {
+		suggestions += "- Review the migration SQL\n"
+		suggestions += "- Check database logs for more details\n"
+	}
+
+	return suggestions
+}
```
avillega: I am unsure about this one in general, but I wanted a way to suggest solutions when the migrations fail. How ever, in practice it is very difficult to correctly handle errors coming from the db given that they are just strings, so tried to use regex for it, but I think it might be brittle.

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,332 @@
...
+	"github.com/resonatehq/resonate/internal/app/subsystems/aio/store/migrations"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	_ "github.com/lib/pq"
+	_ "github.com/mattn/go-sqlite3"
+)
+
+// MigrateConfig holds configuration for the migrate command
+type MigrateConfig struct {
```
dfarr, line 22: To get the config files (as opposed to flags) to work across the serve+migrate commands we need to reuse the same structure like this:
```go
type MigrateConfig struct {
	AIO MigrateAIO `flag:"aio"`
}

type AIO struct {
	Subsystems MigrateAIOSubsystems `flag:"-"`
}

type MigrateAIOSubsystems struct {
	StorePostgres DisabledSubsystem[postgres.Config] `flag:"store-postgres"`
	StoreSqlite   EnabledSubsystem[sqlite.Config]    `flag:"store-sqlite"`
}
```

I recommend doing this using the already defined `postgres.Config` and `sqlite.Config` instead of new structs here so our migrate command picks up the same config options as the serve command.

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,332 @@
...
+
+	return migrateCmd
+}
+
+func newStatusCmd(config *MigrateConfig) *cobra.Command {
+	return &cobra.Command{
+		Use:   "status",
+		Short: "Show current migration status",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			store, err := getMigrationStore(config)
```
dfarr, line 126: Can we move the connection instantiation to the store implementation so that we can be sure the connection is constructed the same way between migrate and serve?

I'm thinking we could add a function like this so we can construct the db without all the extraneous AIO stuff.
```go
func NewConn(config *Config) (*sql.DB, error) {}
```

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,332 @@
...
+			} else {
+				latestVersion = currentVersion
+			}
+
+			pending, err := migrations.GetPendingMigrations(currentVersion, store)
+			if err != nil {
+				return fmt.Errorf("failed to get pending migrations: %w", err)
+			}
+
+			fmt.Printf("Store type: %s\n", store.String())
```
dfarr: Super nitpicky but can we use `cmd.Printf` in the command handlers?

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,332 @@
...
+
+// getMigrationStore infers the store type, opens the database connection,
+// and returns the appropriate migration store. The store owns the database
+// connection and the caller is responsible for calling Close() on the store.
+func getMigrationStore(config *MigrateConfig) (migrations.MigrationStore, error) {
+	// Infer store type from enable flags
+	sqliteEnabled := config.Store.Sqlite.Enabled
+	postgresEnabled := config.Store.Postgres.Enabled
+
+	if postgresEnabled && sqliteEnabled {
```
dfarr: I know this isn't ideal but in serve we let postgres take precedence because the default is sqlite, I'm wondering if we should do the same here? If we throw an error than in order to opt in to postgres you would have to do this:
```
resonate migrate up --aio-store-postgres-enable=true --aio-store-sqlite-enable=false ...
```

internal/app/subsystems/aio/store/migrations/migrations.go
dfarr: Agreed. I think this is almost an impossible task and even if we do it well, an underlying change could break our suggestions. IMO, the risk of providing bad advice is high enough to avoid even trying, I would just print the error itself and leave it at that.

dfarr (changes requested):

internal/app/subsystems/aio/store/migrations/migrations.go
avillega: Perfect, will remove it, and just print the db error as is.

cmd/migrate/migrate.go
avillega: In some way I wanted the serve and migrate commands to not share that many things, their needs are different, for example, you might start postgress with 8 workers for `serve` but you will always want a single worker for `migrate` that is why I also wanted to keep the config separate, since many of the options that apply for `serve` don't make sense or are not necessary for `migrate`

dfarr: You're right, taking a look at the [postgres config](https://github.com/resonatehq/resonate/blob/main/internal/app/subsystems/aio/store/postgres/postgres.go#L386-L398) only a subset of the options are "connection" options. My worry is that the connection logic [here](https://github.com/resonatehq/resonate/blob/main/internal/app/subsystems/aio/store/postgres/postgres.go#L413-L433) could drift from the connection logic in the migration module and we could end up in a situation where the connection works for `serve` but not `migrate` because that codepath is used less frequently.

Could we move some of the logic into a function and pass in the relevant options from both serve and migrate?

avillega: I wont even know what is a good place for that code to live, is it the store code? or like a new file `storeUtils` that have the code for both postgress and sqlite?

dfarr: Couldn't we just have a `NewConn` function in the sqlite/postgres packages?

cmd/migrate/migrate.go
avillega: I kept the config separated but added some test to check that all the relevant properties for migrate are present in the serve config, that way if they diverge we will have an automated way of checking it

cmd/migrate/migrate.go
avillega: yeah, I was overthinking it. Fixed

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,291 @@
...
+type MigrateAIO struct {
+	Subsystems MigrateSubsystems `flag:"-"`
+}
+
+type PersistentEnabledSubsystem[T any] struct {
+	Enabled bool `flag:"enable" desc:"enable subsystem" default:"true" persistent:"true"`
+	Config  T    `flag:"-"`
+}
+
+type PersistenDisabledSubsystem[T any] struct {
```
dfarr:
```suggestion
type PersistentDisabledSubsystem[T any] struct {
```

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,291 @@
...
+	Enabled bool `flag:"enable" desc:"enable subsystem" default:"false" persistent:"true"`
+	Config  T    `flag:"-"`
+}
+
+type MigrateSubsystems struct {
+	StorePostgress PersistenDisabledSubsystem[PostgresConfig] `flag:"store-postgres"`
```
dfarr:
```suggestion
	StorePostgres PersistenrDisabledSubsystem[PostgresConfig] `flag:"store-postgres"`
```

dfarr (approved):
Found some typos, but looks great!

dfarr (approved):

PR #863, opened 2025-10-16
Sender plugin: Kafka

azghr:


dfarr:
Thank you once again @muhammad-asghar-ali !!!

I'm sure you've seen that we're holding off on merging plugin PRs for the time being, however, we have decided that we want one plugin that is (i) durable and (ii) self-hostable and kafka is our natural choice, so if you're still up for it let's work together to get this PR merged!

I noticed this PR is using the sarama kafka library, would it be possible to switch to the [confluent library](https://github.com/confluentinc/confluent-kafka-go)?

azghr:
> Thank you once again @muhammad-asghar-ali !!!
> 
> I'm sure you've seen that we're holding off on merging plugin PRs for the time being, however, we have decided that we want one plugin that is (i) durable and (ii) self-hostable and kafka is our natural choice, so if you're still up for it let's work together to get this PR merged!
> 
> I noticed this PR is using the sarama kafka library, would it be possible to switch to the [confluent library](https://github.com/confluentinc/confluent-kafka-go)?

Yes @dfarr I know that we are holding new plugins util make the independent from main server that why I just created the draft PR for visibility and sure I'll switch from sarama to confluent

dfarr:
> Yes @dfarr I know that we are holding new plugins util make the independent from main server that why I just created the draft PR for visibility and sure I'll switch from sarama to confluent

Thank you!

internal/app/plugins/kafka/kafka.go
```diff
@@ -0,0 +1,192 @@
...
+func New(a aio.AIO, metrics *metrics.Metrics, config *Config) (*Kafka, error) {
+	producer, err := kafka.NewProducer(newConfig(config))
+	if err != nil {
+		return nil, fmt.Errorf("kafka producer: %w", err)
+	}
+	return NewWithProducer(a, metrics, config, producer)
+}
+
+func newConfig(config *Config) *kafka.ConfigMap {
+	c := &kafka.ConfigMap{
```
dfarr: Do you mind just moving this up to line 62, I don't think we need a function for this

internal/app/plugins/kafka/kafka.go
```diff
@@ -0,0 +1,192 @@
...
+		"bootstrap.servers":   fmt.Sprintf("%v", config.Brokers),
```
dfarr: I'm wondering if this is supposed to be in the format of `"http://broker1.com, http://broker2.com, ..."`. If that is the case can we use `strings.Join`?

internal/app/plugins/kafka/kafka.go
```diff
@@ -0,0 +1,192 @@
...
+		"delivery.timeout.ms": int(config.Timeout.Milliseconds()),
```
dfarr: I wonder if it's possible to convert the int64 here to a string instead of an int?

go.mod
```diff
@@ -4,8 +4,10 @@ go 1.24.0
 
 require (
 	github.com/anishathalye/porcupine v1.0.3
+	github.com/aws/aws-sdk-go-v2 v1.39.4
 	github.com/aws/aws-sdk-go-v2/config v1.31.15
 	github.com/aws/aws-sdk-go-v2/service/sqs v1.42.11
+	github.com/confluentinc/confluent-kafka-go v1.9.2
```
dfarr: It looks like there is a v2 version of this library, can we upgrade to the latest release?

internal/app/plugins/kafka/kafka.go
```diff
@@ -0,0 +1,192 @@
...
+		msg.Headers = make([]kafka.Header, 0, len(addr.Headers))
+		for k, v := range addr.Headers { // nosemgrep: range-over-map
+			msg.Headers = append(msg.Headers, kafka.Header{
+				Key:   k,
+				Value: []byte(v),
+			})
+		}
+	}
+
+	err := w.producer.Produce(msg, nil)
```
dfarr: Can you add a deliverChan argument to the [call to produce](https://docs.confluent.io/platform/current/clients/confluent-kafka-go/index.html#Producer.Produce) so we can wait for acknowledgment?

dfarr: We will need to wait for a message to be sent to this channel, and according to the docs
> `*kafka.Message` - delivery report for produced message. Check `.TopicPartition.Error` for delivery result.

internal/app/plugins/kafka/kafka.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	w.producer.Flush(1000)
```
dfarr: I think if we wait for a message on the delivery chan we don't need to call Flush

dfarr (changes requested):
Thank you once again @muhammad-asghar-ali! Let me know if you have any follow up questions :)

internal/app/plugins/kafka/kafka.go
azghr: done remove the function

internal/app/plugins/kafka/kafka.go
azghr: resolved

internal/app/plugins/kafka/kafka.go
azghr: done converted int64 to string

go.mod
azghr: done upgrade the package version

internal/app/plugins/kafka/kafka.go
azghr: delivery handling now waits on the producer’s delivery channel, so there’s no manual Flush call.

internal/app/plugins/kafka/kafka.go
azghr: done and removed the manual Flush function

dfarr:
Thank you so much @muhammad-asghar-ali ! Do you mind updating this branch and maybe we can use your base plugin here too?

azghr:
> Thank you so much @muhammad-asghar-ali ! Do you mind updating this branch and maybe we can use your base plugin here too?

Introduced the base logic in kafka plugin

dfarr:
Completed by #930

PR #890, opened 2025-10-22
Sender plugin: Base plugin architecture

azghr:
- Migrate HTTP and SQS plugins to use base.Plugin for common functionality
- Embed base.BaseConfig in all plugin configs for consistent structure  
- Create processor interfaces (HttpProcessor → processor, SQSProcessor → processor)
- Update parameter names (addr → data) for consistency with base plugin
- Keep Poll plugin's original architecture due to complex connection management
- Update all tests to use new BaseConfig structure
- Reduce code duplication across plugins while maintaining backward compatibility

### DIFF
--- a/internal/app/plugins/base/base.go
+++ b/internal/app/plugins/base/base.go
@@ -1,0 +1,133 @@
+package base
+
+import (
+	"fmt"
+	"log/slog"
+	"strconv"
+	"time"
+
+	"github.com/resonatehq/resonate/internal/aio"
+	"github.com/resonatehq/resonate/internal/kernel/t_aio"
+	"github.com/resonatehq/resonate/internal/metrics"
+)
+
+type BaseConfig struct {
+	Size        int           `flag:"size" desc:"submission buffered channel size" default:"1000"`
+	Workers     int           `flag:"workers" desc:"number of workers" default:"4"`
+	Timeout     time.Duration `flag:"timeout" desc:"request timeout" default:"30s"`
+	TimeToRetry time.Duration `flag:"ttr" desc:"time to wait before resending" default:"15s"`
+	TimeToClaim time.Duration `flag:"ttc" desc:"time to wait for claim before resending" default:"1m"`
+}
+
+type Processor interface {
+	Process(body []byte, addr []byte) (bool, error)
+}
+
+type Worker struct {
+	id        int
+	sq        <-chan *aio.Message
+	processor Processor
+	config    *BaseConfig
+	aio       aio.AIO
+	metrics   *metrics.Metrics
+	name      string
+}
+
+type Plugin struct {
+	name    string
+	sq      chan *aio.Message
+	workers []*Worker
+	cleanup func() error
+}
+
+func NewPlugin(a aio.AIO, name string, config *BaseConfig, metrics *metrics.Metrics, processor Processor, cleanup func() error) *Plugin {
+	sq := make(chan *aio.Message, config.Size)
+	workers := make([]*Worker, config.Workers)
+
+	for i := 0; i < config.Workers; i++ {
+		workers[i] = &Worker{
+			id:        i,
+			sq:        sq,
+			processor: processor,
+			config:    config,
+			aio:       a,
+			metrics:   metrics,
+			name:      name,
+		}
+	}
+
+	return &Plugin{
+		name:    name,
+		sq:      sq,
+		workers: workers,
+		cleanup: cleanup,
+	}
+}
+
+func (p *Plugin) String() string {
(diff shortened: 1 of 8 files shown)


internal/app/plugins/http/http.go
```diff
@@ -3,128 +3,33 @@ package http
...
 	"github.com/resonatehq/resonate/internal/metrics"
 )
 
 type Config struct {
-	Size        int           `flag:"size" desc:"submission buffered channel size" default:"100"`
-	Workers     int           `flag:"workers" desc:"number of workers" default:"1"`
-	Timeout     time.Duration `flag:"timeout" desc:"http request timeout" default:"1s"`
-	TimeToRetry time.Duration `flag:"ttr" desc:"time to wait before resending" default:"15s"`
-	TimeToClaim time.Duration `flag:"ttc" desc:"time to wait for claim before resending" default:"1m"`
+	base.BaseConfig
```
dfarr: Could you do me a favour and not embed the baseConfig? The reason is that the default values differ across the plugins and these values are specified in the struct tag. Using an embedded struct means we can not set the default size, workers, etc differently for each plugin.

You can see what the defaults are currently set to by running:
```
go run ./... serve --help
```

internal/app/plugins/base/base.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/resonatehq/resonate/internal/aio"
+	"github.com/resonatehq/resonate/internal/kernel/t_aio"
+	"github.com/resonatehq/resonate/internal/metrics"
+)
+
+type BaseConfig struct {
+	Size        int           `flag:"size" desc:"submission buffered channel size" default:"1000"`
+	Workers     int           `flag:"workers" desc:"number of workers" default:"4"`
+	Timeout     time.Duration `flag:"timeout" desc:"request timeout" default:"30s"`
```
dfarr: I don't think the base config needs timeout

internal/app/plugins/poll/poll.go
```diff
@@ -14,22 +14,21 @@ import (
...
 	"github.com/resonatehq/resonate/internal/aio"
+	"github.com/resonatehq/resonate/internal/app/plugins/base"
 	"github.com/resonatehq/resonate/internal/kernel/t_aio"
 	"github.com/resonatehq/resonate/internal/metrics"
 	"github.com/resonatehq/resonate/internal/util"
 )
 
 type Config struct {
-	Size            int               `flag:"size" desc:"submission buffered channel size" default:"100"`
+	base.BaseConfig
```
dfarr: Once we unwind embedding for the poller plugin we will need to be careful to hardcode the number of workers in the base config to 1. Due to the design of the poller plugin it is important that there is a single worker only.

internal/app/plugins/sqs/sqs.go
```diff
@@ -4,155 +4,52 @@ import (
...
+	"github.com/resonatehq/resonate/internal/app/plugins/base"
 	"github.com/resonatehq/resonate/internal/metrics"
 )
 
 type Config struct {
-	Size        int           `flag:"size" desc:"submission buffered channel size" default:"100"`
-	Workers     int           `flag:"workers" desc:"number of workers" default:"1"`
-	Timeout     time.Duration `flag:"timeout" desc:"aws request timeout" default:"30s"`
-	TimeToRetry time.Duration `flag:"ttr" desc:"time to wait before resending" default:"15s"`
-	TimeToClaim time.Duration `flag:"ttc" desc:"time to wait for claim before resending" default:"0"`
```
dfarr, line 26: For sqs it is important that we keep the default TimeToClaim as 0. Because sqs is a durable queue, once a message has been published to sqs we never attempt to resend the message by default (indicated by a 0).

dfarr (changes requested):
Thank you once again @muhammad-asghar-ali! My only request is to unwind the embedded struct. It is important that each plugin maintains it's own config as opposed to "inheriting" from a base config.

internal/app/plugins/http/http.go
azghr: done

internal/app/plugins/base/base.go
azghr: done

internal/app/plugins/sqs/sqs.go
azghr: done

internal/app/plugins/poll/poll.go
azghr: revert to old poll config struct

dfarr (approved):
Thank you @muhammad-asghar-ali !!!

avillega (commented):
Looks good, very clean

PR #900, opened 2025-11-04
Add promise, schedule, and task metrics

dfarr:
Adds the following metrics:
- promises{created, resolved, rejected, canceled, timedout}
- schedules{created, deleted}
- tasks{created, dropped, completed, timedout}

### DIFF
--- a/internal/app/coroutines/claimTask.go
+++ b/internal/app/coroutines/claimTask.go
@@ -17,13 +17,10 @@ import (
 func ClaimTask(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any], r *t_api.Request) (*t_api.Response, error) {
 	req := r.Payload.(*t_api.ClaimTaskRequest)
 
+	config, ok := c.Get("config").(*system.Config)
+	util.Assert(ok, "coroutine must have config dependency")
 	util.Assert(req.ProcessId != "", "process id must be set")
 	util.Assert(req.Ttl >= 0, "ttl must be greater than or equal to 0")
-
-	config, ok := c.Get("config").(*system.Config)
-	if !ok {
-		panic("coroutine must have config dependency")
-	}
 
 	var status t_api.StatusCode
 	var t *task.Task
--- a/internal/app/coroutines/claimTask.go
+++ b/internal/app/coroutines/claimTask.go
@@ -16,11 +16,10 @@ import (
 
 func ClaimTask(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any], r *t_api.Request) (*t_api.Response, error) {
 	req := r.Payload.(*t_api.ClaimTaskRequest)
-
-	config, ok := c.Get("config").(*system.Config)
-	util.Assert(ok, "coroutine must have config dependency")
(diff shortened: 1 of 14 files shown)


dfarr:
This PR looks bigger than it is because I refactored the background coroutines [here](https://github.com/resonatehq/resonate/pull/900/files#diff-9e9b67c40b251bf355731ba916c3e59e683c3bf901ef6da398e637677b048caeR44). This normalizes request-based coroutines and background coroutines to retrieve config and metrics from the dependencies. This affects the following coroutines:
- TimeoutPromises
- SchedulePromises
- TimeoutLocks
- EnqueueTasks
- TimeoutTasks

internal/app/coroutines/claimTask.go
```diff
@@ -17,14 +17,11 @@ import (
 func ClaimTask(c gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any], r *t_api.Request) (*t_api.Response, error) {
 	req := r.Payload.(*t_api.ClaimTaskRequest)
 
+	config, ok := c.Get("config").(*system.Config)
+	util.Assert(ok, "coroutine must have config dependency")
```
avillega: The `,ok` might be unnecessary here since we are doing a type assertion anyway, if the "config" is not present the program will panic trying to convert null into `system.Config`, applies to getting the "metrics" too.

internal/kernel/system/system.go
```diff
@@ -209,10 +212,21 @@ func (s *System) AddOnRequest(kind t_api.Kind, constructor func(gocoro.Coroutine
 	}
 }
 
-func (s *System) AddBackground(name string, constructor func(*Config, map[string]string) gocoro.CoroutineFunc[*t_aio.Submission, *t_aio.Completion, any]) {
+func (s *System) AddBackground(name string, constructor func(gocoro.Coroutine[*t_aio.Submission, *t_aio.Completion, any], map[string]string) (any, error)) {
```
avillega, line 215: This is nice!

internal/app/coroutines/timeoutPromises.go
avillega, line 1: I wonder if we should have a metric for completed and timed out promises, we do have one for created promises. Would be nice to see that the created promises are being resolved.

avillega (approved):

internal/kernel/system/system.go
dfarr: Thank you!

internal/app/coroutines/timeoutPromises.go
dfarr: We have this [here](https://github.com/resonatehq/resonate/pull/900/files#diff-3564a39358c73c5cc39d94d319cb43f3657b026b3b7bf706fd316a6e71767debR210). We are counting resolved, rejected, canceled and timedout separately. From my understanding this is the idiomatic way to do because you can derive completed and pending (which is why I call the state "created" instead of pending):
```
completed = resolved + rejected + canceled + timedout
pending = created - completed
```

dfarr: But I'm not super knowledgeable on these metrics, if it makes more sense to maintain a "completed" counter and a pending gauge separately that would be relatively easy to do in the future.

PR #926, opened 2025-11-18
Set up some infrastructure for auth and auz

avillega:
This is ready for review. How ever this is not the whole feature yet.
This PR introduces the concept of a Middleware at the API layer, this middleware are functions that take a `*t_api.Request` and might return an error, if an error is returned when executing the any of the middlewares, the error is returned.
The JWTAuthenticator is a middleware

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -34,19 +34,21 @@ type API interface {
 // API
 
 type api struct {
-	sq         chan *bus.SQE[t_api.Request, t_api.Response]
-	buffer     *bus.SQE[t_api.Request, t_api.Response]
-	subsystems []Subsystem
-	done       bool
-	errors     chan error
-	metrics    *metrics.Metrics
+	sq            chan *bus.SQE[t_api.Request, t_api.Response]
+	buffer        *bus.SQE[t_api.Request, t_api.Response]
+	subsystems    []Subsystem
+	done          bool
+	errors        chan error
+	metrics       *metrics.Metrics
+	authenticator Authenticator
 }
 
-func New(size int, metrics *metrics.Metrics) *api {
+func New(size int, metrics *metrics.Metrics, authenticator Authenticator) *api {
 	return &api{
-		sq:      make(chan *bus.SQE[t_api.Request, t_api.Response], size),
-		errors:  make(chan error),
-		metrics: metrics,
+		sq:            make(chan *bus.SQE[t_api.Request, t_api.Response], size),
+		errors:        make(chan error),
+		metrics:       metrics,
+		authenticator: authenticator,
 	}
 }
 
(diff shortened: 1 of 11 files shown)


internal/api/api.go
```diff
@@ -183,6 +188,14 @@ func (a *api) EnqueueSQE(sqe *bus.SQE[t_api.Request, t_api.Response]) {
 		return
 	}
 
+	// Run all the middlewares
+	for _, m := range a.middlewares {
+		err := m.Process(sqe.Submission)
+		if err != nil {
+			sqe.Callback(nil, err)
```
dfarr, line 195: I think we should return here

internal/api/middleware.go
```diff
@@ -0,0 +1,187 @@
...
+func extractPromiseId(taskId string) string {
+	// We expect the taskId to have the one of following formats
+	// __resume:{promiseId}:{another}
+	// __notify:{promiseId}:{another}
+	// __invoke:{promiseId}
+	// if that changes we need to change this code
+	if !strings.HasPrefix(taskId, "__resume") &&
+		!strings.HasPrefix(taskId, "__invoke") &&
+		!strings.HasPrefix(taskId, "__notify") {
+		panic("taskId must start with __resume, __invoke, or __notify")
```
dfarr: I don't think we should panic here because I can specify any id I want in the request, and the same goes for the assertion below.

Parsing the string makes me a little nervous because if we do it wrong it could be a security hole.  What if we could do it without parsing like this?

```go
 strings.HasPrefix(taskId, fmt.Sprintf("__invoke:%s", id)) || 
 strings.HasPrefix(taskId, fmt.Sprintf("__resume:%s", id)) || 
 strings.HasPrefix(taskId, fmt.Sprintf("__notify:%s", id))
```

internal/app/subsystems/api/http/http.go
```diff
@@ -28,6 +28,7 @@ import (
 type Config struct {
 	Addr          string            `flag:"addr" desc:"http server address" default:":8001"`
 	Auth          map[string]string `flag:"auth" desc:"http basic auth username password pairs"`
+	PublicKeyPath string            `flag:"auth-public-key" desc:"public key path used for jwt based authentication"`
```
dfarr: I wonder if we should put this on the [API config](https://github.com/resonatehq/resonate/blob/main/cmd/config/config.go#L58-L61) instead? Could this apply to different api implementations in the future?

internal/kernel/t_api/status.go
```diff
@@ -51,6 +53,10 @@ func (s StatusCode) String() string {
 		return "The request was successful"
 	case StatusFieldValidationError:
 		return "The request is invalid"
+	case StatusUnauthorized:
+		return "The request is missing valid authentication credentials"
```
```suggestion
		return "The request is unauthorized"
```

dfarr (changes requested):
This looks really really good. As simple as I hoped in a best case scenario!

One annoying typo: the plural of middleware is middleware (sort of like how the plural of software is software).

internal/app/subsystems/api/http/http.go
avillega: That is the idea, however I am starting with http only for now, need to go deeper into gRPC to understand how they do metadata and have no idea how this will look for any other api subsystem like kafka

dfarr: I'm wondering if we should still put the auth config on the API as this way it won't require a backwards incompatible change later. 

We can always warn or even panic if auth is enabled with any other API subsystem.

avillega: At the moment the request will be rejected because it will not contain the bearer token.

dfarr (approved):
🎉

PR #940, opened 2025-12-01
Http support for jwt auth and auz

avillega:


avillega:
Manual tests:
 - Generate a public/private key pair using openssl
 - using a tool like `jwt-cli` generate a jwt token with RSA256 (only one supported at the moment) signature, use the private key to sign token
```
# example command for a token with access to the prefix foo:
jwt encode -A RS256 -S @private_key.pem -P prefix='foo'
```
 - start the server with the public key
```
# example command:
./re serve --api-auth-public-key public_key.pem
```
 - make request using the generate token
```
example commands:
resonate promises create foo --timeout 1h --token "{the genearted token here}" # should succeed
resonate promises create bar --timeout 1h --token "{the genearted token here}" # should fail unauthorized prefix
resonate promises create foo --timeout 1h # should fail (no token
```

dfarr (approved):
🎉

PR #959, opened 2025-12-10
Clamp int64 additions

dfarr:
A ttl value added to the current timestamp (in ms) causes int64 overflow, guard against this.

The python task transition tests use a ttl of `sys.maxsize` which causes integer overflow. This problem was masked likely by the system tick frequency but recent changes exposed this problem and the tests are now failing against the remote store. This bug fix fixes the tests and is a good idea anyways.

### DIFF
--- a/cmd/util/flags.go
+++ b/cmd/util/flags.go
@@ -124,30 +124,6 @@ func (f *rangeFlag[T]) Resolve(r *rand.Rand) T {
 	return f.resolve(f, r)
 }
 
-func RangeIntn(r *rand.Rand, min int, max int) int {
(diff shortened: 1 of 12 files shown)


avillega (commented):
Should we change the DST generator to use sys.max (or the qeuivalent in go) ttl to test this does not break again in the future?

dfarr:
DST runs:
- https://github.com/resonatehq/resonate/actions/runs/20281393283
- https://github.com/resonatehq/resonate/actions/runs/20281390113

avillega (approved):

PR #993, opened 2026-01-06
Remove strict and ikey

avillega:


Tomperez98 (approved):
Next steps:
- Remove iKey and strict from the SDK
- Modify the autogenerated tests to account for this change

dfarr (approved):

avillega:
Confirmed this is a non breaking change for the sdk, current version of the sdk works with this change

PR #1002, opened 2026-01-13
new tree command

Tomperez98:
New format (with resonate:origin):

1 initial search + pagination calls (if the result set is large)
Gets the entire call graph in one query
Number of API calls = 1 + (number of pagination rounds)

Old format (with resonate:root):

1 search per branch in the call graph
If you have a call graph with N branches (global scope promises), you need N separate searches
Each search may also require pagination
Number of API calls = N branches × (1 + pagination rounds per branch)

Example:
If you have a call graph with 5 branches and each branch requires 2 pagination rounds:

New format: 1 + 2 = 3 API calls total
Old format: 5 × (1 + 2) = 15 API calls total

The new format is more efficient because resonate:origin represents the entire call graph ID, so all promises in that graph share the same origin tag. The old format required traversing the tree structure and making separate API calls for each branch because resonate:root only represented individual branch roots, not the entire graph.

### DIFF
--- a/cmd/tree/tree.go
+++ b/cmd/tree/tree.go
@@ -15,7 +15,7 @@ import (
 
 type searchParams struct {
 	id     string
-	root   string
+	origin string
 	cursor *string
 }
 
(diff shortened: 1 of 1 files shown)


cmd/tree/tree.go
```diff
@@ -48,45 +48,99 @@ func NewCmd() *cobra.Command {
...
 				}
+			}
 
-				if res.JSON200.Cursor != nil {
-					searches = append(searches, &searchParams{
-						id:     params.id,
-						root:   params.root,
-						cursor: res.JSON200.Cursor,
+			// Fallback to old format if new format returned no results
+			if !usingNewFormat {
```
dfarr: Can we make this an else clause?

cmd/tree/tree.go
```diff
@@ -48,45 +48,99 @@ func NewCmd() *cobra.Command {
...
+
+					// Handle pagination
+					if res.JSON200.Cursor == nil {
+						break
 					}
 
-					if p.Tags["resonate:scope"] == "global" && p.Id != params.root {
-						searches = append(searches, &searchParams{id: "*", root: p.Id})
+					res, err = c.V1().SearchPromisesWithResponse(context.TODO(), &v1.SearchPromisesParams{
+						Id:     &[]string{"*"}[0],
```
dfarr: Mind just make this a string pointer?

cmd/tree/tree.go
```diff
@@ -48,45 +48,99 @@ func NewCmd() *cobra.Command {
 			}
 
 			var rootPromise *v1.Promise
-			searches := []*searchParams{{id: "*", root: args[0]}}
 			promises := map[string][]*v1.Promise{}
+			usingNewFormat := false
+
+			// Try new tag format first (resonate:origin)
+			res, err := c.V1().SearchPromisesWithResponse(context.TODO(), &v1.SearchPromisesParams{
+				Id:   &[]string{"*"}[0],
```
dfarr: Mind just make this a string pointer?

dfarr (changes requested):

dfarr (approved):

PR #1009, opened 2026-01-20
Single endpoint api

avillega:


internal/app/subsystems/api/http/process.go
```diff
@@ -0,0 +1,413 @@
...
+	Pid     string `json:"pid" binding:"required"`
+	Ttl     int64  `json:"ttl" binding:"required"`
+}
+
+type taskReleaseData struct {
+	Id      string `json:"id" binding:"required"`
+	Version int    `json:"version" binding:"required"`
+}
+
+type taskHeartbeatData struct {
```
dfarr, line 96: We have extended task heartbeat to account for per-task heartbeating, [see here ](https://github.com/resonatehq/protocol-specification?tab=readme-ov-file#heartbeat).

Should we account for this now? In phase 1 we would include an empty list of tasks and fallback to heartbeating by pid.

dfarr (approved):
Looks good! The suggestion regarding heartbeating might not be necessary to implement since extraneous fields (list of tasks in this case) are probably ignored by default.

dfarr (approved):

PR #1012, opened 2026-01-26
Task.Fulfill

avillega:


internal/app/coroutines/taskFulfill.go
```diff
@@ -0,0 +1,212 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// NOTE: if the the completion of the task was not applied it means the task
+		// got moved from acquired to another state while doing the current operation
+		// return 409
+
+		if !ok {
+			slog.Debug("Couldn't fulfill task and settle promise", "taskId", t.Id, "version", t.Counter, "promiseId", p.Id)
```
dfarr, line 83: Could we re-run the coroutine in this case?

internal/app/coroutines/taskFulfill.go
```diff
@@ -0,0 +1,212 @@
...
+			Data:   &t_api.TaskFulfillResponse{},
+		}, nil
+	}
+
+	p, err := gocoro.SpawnAndAwait(c, readPromise(r.Head, req.Action.Id))
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: TaskV2, assertion promise must exists
```
dfarr, line 54: In taskv2 this would be an invariant violation, I almost wonder if we should panic

dfarr (approved):

internal/app/coroutines/taskFulfill.go
avillega: Yes, in taskV2 this will be a panic

PR #1013, opened 2026-01-27
Actually task.suspend

avillega:


internal/kernel/t_api/request.go
```diff
@@ -358,6 +359,43 @@ func (r *TaskFulfillRequest) Kind() Kind {
 	return TaskFulfill
 }
 
+type TaskSuspendRequest struct {
+	Id      string `json:"id"`
+	Version int    `json:"version"`
+	Actions []PromiseRegisterRequest
```
dfarr, line 365: Does this need a struct tag?

dfarr (approved):

internal/kernel/t_api/request.go
avillega: At the moment we never automatically marshall or unmarshall this struct, thats why I didn't add it

PR #1108, opened 2026-05-25
Make http push and gcp non-blocking

### LINKED ISSUE #1067: Non-blocking message dispatch with per-transport concurrency caps
## Problem

The message processing loop in `src/processing/processing_messages.rs` dispatches execute and unblock messages sequentially via two `dispatcher.send(...)` calls inside `process_batch`. Both calls are `.await`ed, so the loop blocks on each send before moving to the next message.

For the HTTP push transport, this means a single slow or unresponsive webhook can stall the loop for up to 180 seconds (the current request timeout), delaying all subsequent messages in the batch.

GCP Pub/Sub already spawns a `tokio::spawn` internally but has no concurrency cap and a hardcoded 10s timeout.  
HTTP poll uses `try_send` on an in-memory channel and is already non-blocking — no changes needed there.

## Solution

Make `dispatcher.send` fire-and-forget by spawning a tokio task per message, bounded by a per-transport semaphore. If the semaphore is exhausted, log a warning and drop the message rather than blocking the loop.

### HTTP Push (`src/transport/transport_http_push.rs`)

- Add an `Arc<Semaphore>` to `HttpPushTransport`, sized from the existing (but currently unwired) `concurrency` config field (`src/config.rs:341`).
- In `send`, call `semaphore.try_acquire_owned()`:
  - If it fails → `tracing::warn!` and return immediately (message dropped).
  - If it succeeds → `tokio::spawn` the HTTP request; the `OwnedSemaphorePermit` is held by the spawned task and released on completion.
- `dispatcher.send` returns immediately in both cases; the loop is no longer blocked.

### GCP Pub/Sub (`src/transport/transport_gcps.rs`)

- Apply the same semaphore pattern.
- Move the hardcoded `10s` timeout into config.

### Config & CLI changes

**HTTP Push** — config and CLI flags already exist, just need to be wired to the semaphore:

| CLI Flag | Env Var | Default |
|---|---|---|
| `--transports-http-push-concurrency N` | `RESONATE_TRANSPORTS__HTTP_PUSH__CONCURRENCY` | 16 |
| `--transports-http-push-connect-timeout MS` | `RESONATE_TRANSPORTS__HTTP_PUSH__CONNECT_TIMEOUT` | 10000 |
| `--transports-http-push-request-timeout MS` | `RESONATE_TRANSPORTS__HTTP_PUSH__REQUEST_TIMEOUT` | 180000 |

**GCP Pub/Sub** — new fields needed in `GcpsConfig` (`src/config.rs`) and `CommonArgs` (`src/cli.rs`):

| CLI Flag | Env Var | Default |
|---|---|---|
| `--transports-gcps-concurrency N` | `RESONATE_TRANSPORTS__GCPS__CONCURRENCY` | 16 |
| `--transports-gcps-timeout MS` | `RESONATE_TRANSPORTS__GCPS__TIMEOUT` | 10000 |

## Files to change

- `src/config.rs` — add `concurrency` and `timeout` to `GcpsConfig`
- `src/cli.rs` — add `--transports-gcps-concurrency` and `--transports-gcps-timeout` to `CommonArgs`; wire new fields in `apply`
- `src/transport/transport_http_push.rs` — add semaphore, spawn task, drop on exhaustion
- `src/transport/transport_gcps.rs` — add semaphore, move timeout to config
- `src/main.rs` — pass concurrency values when constructing transports

## Acceptance criteria

- [ ] `process_batch` returns without waiting for any network I/O
- [ ] HTTP push sends are capped at `concurrency` concurrent in-flight requests; excess messages are dropped with a warning log
- [ ] GCP Pub/Sub sends are capped at `concurrency` concurrent in-flight publishes; excess messages are dropped with a warning log
- [ ] GCP Pub/Sub timeout is driven by config, not hardcoded
- [ ] All four new/existing flags are accepted via CLI and env var
- [ ] HTTP poll behaviour is unchanged

Tomperez98:


dfarr (approved):

src/transport/transport_gcps.rs
```diff
@@ -3,92 +3,152 @@
...
-        }
-    }
+    pub fn new(concurrency: usize, timeout: Duration) -> Self {
+        let publishers = Arc::new(Mutex::new(HashMap::<String, Publisher>::new()));
+        let semaphore = Arc::new(Semaphore::new(concurrency));
+        // Queue capacity larger than concurrency so short bursts smooth out;
+        // full queue + full in-flight pushes back on `send()`, which in turn
+        // pushes back on the claim loop in processing_messages — the DB is
+        // the durable buffer.
+        let (tx, rx) = mpsc::channel::<PublishJob>(64 * concurrency);
```
dfarr: should we just go with concurrency?

src/transport/transport_http_push.rs
```diff
@@ -118,61 +119,125 @@ impl Auth {
...
             .build()
             .expect("failed to build HTTP client");
-        Self { client, auth }
+        let auth = Arc::new(auth);
+        let semaphore = Arc::new(Semaphore::new(concurrency));
+        // Queue capacity is intentionally larger than `concurrency` so short
+        // bursts smooth out; full queue + full in-flight pushes back on
+        // `send()`, which in turn pushes back on the claim loop in
+        // processing_messages — the DB is the durable buffer.
+        let (tx, rx) = mpsc::channel::<DeliveryJob>(64 * concurrency);
```
dfarr: same here

dfarr (approved):

